PR #3, opened 2021-06-16
Loadout fixes lost in the repo move

MrMelbert:
I fixed this in the past, but at some point between the upstream merge and repo move it was unfixed.

Same fix, also some added sanity so it fail less silently.


Jolly-66 (changes requested):
code/game/objects/items/plushes.dm
 Add it to the readme in jollystation_modules

Jolly-66 (approved):

PR #27, opened 2021-07-05
Updates everything to be correctly +15 years from /tg/ time.

CRITAWAKETS:
## About The Pull Request

Every instance of the +540 year increment is updated to the +555 increment now, with a note on it being a non-module change.

## Why It's Good For The Game

The time is now accurate, and we can tell what time we actually are now.

Also the better solution to this would be to change the time increment to a config option, but that's a refactor for /tg/ to do.

### DIFF
--- a/code/game/machinery/computer/medical.dm
+++ b/code/game/machinery/computer/medical.dm
@@ -490,7 +490,7 @@
 				var/counter = 1
 				while(active2.fields[text("com_[]", counter)])
 					counter++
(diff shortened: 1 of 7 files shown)


Jolly-66 (changes requested):
Update the README.md in /jollystation_modules/ so these changes are documented
Preferably in descending order as they would appear in the .dme

Jolly-66 (approved):

PR #29, opened 2021-07-08
Pain expansion part two

MrMelbert:
- Added shock blankets. Shock blankets are found in emergency medkits and in medical vendors. Covering a person in a shock blanket will regulate their body temperature to safe levels while they're un-moving and lying down / buckled to something. They're best used to prevent the temperature loss due to shock (obviously), but can be used if the environment is too cold or too hot in an emergency. 
- Added general painkiller medipens to emergency medkits. These heal about 20% pain very quickly, and another 10%+ over a longer amount of time.
- Added temperature packs. They're found in medical vendors. When activated, you can apply temperature packs to hurt limbs to help heal them over time. Hotter temperature packs have a better pain modifier on the  limb, while colder packs heals pain at a flat rate faster. Temperature packs slow you down while pressing them against a limb, and you can press them against other people, too.
- Added a temperature pack element. Anything can be made into a temperature pack by applying this bespoke element. Currently bottles and things frozen by freon work as cold temperature packs.
- Added sharp pain. Sharp pain is for pain that is much more sudden and extreme but expires quicker. Headaches, getting shocked, and falling from a z-level can cause sharp pain.
- Lying down / sitting down now helps against pain.
- Being on stasis is no longer a x0 pain modifier and instead stops natural pain decay.
- Refactored temporary minimum pain into a status effect.
- Changed our span defines to use concatenation.


jollystation_modules/code/__DEFINES/spans.dm
```diff
@@ -1,8 +1,8 @@
 // -- Modular spans and fonts.
 
-#define span_message(str) ("<span class='message'>[str]</span>")
-#define span_italics(str) ("<span class='italics'>[str]</span>")
-#define span_readable_yellow(str) ("<font color = '#c5c900'>[str]</font>")
+#define span_message(str) ("<span class='message'>" + str + "</span>")
+#define span_italics(str) ("<span class='italics'>" + str + "</span>")
```
Jolly-66, line 4: don't these need to be converted? or are these fine because they're defines?

Jolly-66: going through i'm assuming using span_whatever is only for to chat of carbon mobs

MrMelbert: i don't know what you're talking about, span_[x] is for everything, i'm converting them to the correct style in this PR

prodirus (commented):
mostly seems alright, just minor typos

Jolly-66 (approved):
prod seems content, and the latests commits seem to be mostly commets/grammar stuff 
all good here

PR #38, opened 2021-07-16
July 15th Upstream Merge

Jolly-66:
Shit compiles but since you assholes voted Tram I kept crashing out of VSCode.

Seems good, if we run into issues I can always redo a merge this weekend.


MrMelbert (commented):
first look-over seems good to me, skell tentacles need to be converted to mutant organs now but that can wait

PR #40, opened 2021-07-17
Fixes Skrell again, ports their head tentacles to the new external organ feature, documents their files better

MrMelbert:
image
- Skrell Head Tentacles work again.
- Skrell Head Tentacles are now an organ, like lizard frills/snout, moth wings/antenna. 
- They also work in DNA editing, too
- Doc'd their files better and actually added most of them to the readme
- Added species whitelists and blacklists for quirks. Certain quirks can only be taken if you're a certain species, and certain quirks cannot be taken if you're a certain species. This was added so Skrell can't have light drinker, since they are innately light drinkers.


Jolly-66 (commented):
With the mutant organs bullshit, does it take the head tentacles overlay from the dmi and uses it as an organ Sprite?

MrMelbert:
> 
> 
> With the mutant organs bullshit, does it take the head tentacles overlay from the dmi and uses it as an organ Sprite?

The organ itself has no sprite, but if it exists in a mob and fulfills the `can_draw_on_bodypart` check, it adds an overlay to the limb / mob it occupies.

PR #45, opened 2021-07-20
Painkillers: Miner pen buff, emergency kit pen perf

MrMelbert:
- The painkiller survival pen now heals slightly less pain than it did before, but addicts you 3x slower, which means you can take 4-5 pens before getting addicted now.
- The emergency painkiller pen now heals slightly more pain but addicts you 3x faster. Using 2 emergency painkiller pens will addict you now, as they're intended... for emergencies.
- aspirin / paracetamol / caffeine now has a pain modifier like the other painkillers.


Jolly-66:
what do we tell larry bros

prodirus (approved):

PR #52, opened 2021-07-27
Adds in cybernetic cat ears.

CRITAWAKETS:
Adds in cybernetic cat ears. Nuff said. They require just the basic cybernetic organ tech and have a lower damage modifier alongside being blue.


Jolly-66:
I think somewhere later we'll move these into "Synthetic Mutant Organs" as a tech node since I want to add more of these in the future, namely lizard tongues.

jollystation_modules/code/modules/research/designs/medical_designs.dm
```diff
@@ -0,0 +1,10 @@
+/datum/design/cybernetic_ears/cat
+	name = "Cybernetic Cat Ears"
+	desc = "A pair of cybernetic cat ears."
+	id = "cybernetic_cat_ears"
+	build_type = PROTOLATHE | AWAY_LATHE | MECHFAB
```
Jolly-66, line 5: `AWAY_LATHE`? Isn't that for space ruins or whatever (like Charlie?, and not the Derelict?)

CRITAWAKETS: its just to say that it can be built by ancient protolathes, like basically any protolathe item but certain things like GPS devices

MrMelbert (approved):
The addition to the tech-web can be done slightly better (via `/New()`) to prevent potential future problems, but for now works for me

PR #53, opened 2021-07-28
Changes some of the UI buttons

Jolly-66:
This swaps around some of the buttons + adds in our own things;
Modular CL (which links to the doc), our rules (Discord link, must be logged into Discord desktop). Also moves the buttons around and stuff. If we can, we could set up our own CL like /TG/ has.

Housekeeping: Added myself to the config folder as a code owner.

image

### DIFF
--- a/interface/interface.dm
+++ b/interface/interface.dm
@@ -1,7 +1,7 @@
 //Please use mob or src (not usr) in these procs. This way they can be called in the same fashion as procs.
 /client/verb/wiki(query as text)
 	set name = "wiki"
-	set desc = "Type what you want to know about.  This will open the /TG/ wiki in your web browser. Type nothing to go to the main page."
+	set desc = "Type what you want to know about.  This will open the /tg/station wiki in your web browser. Type nothing to go to the main page." //NON-MODULE CHANGE
 	set hidden = TRUE
 	var/wikiurl = CONFIG_GET(string/wikiurl)
 	if(wikiurl)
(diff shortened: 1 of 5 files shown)


interface/interface.dm
```diff
@@ -1,7 +1,7 @@
 //Please use mob or src (not usr) in these procs. This way they can be called in the same fashion as procs.
 /client/verb/wiki(query as text)
 	set name = "wiki"
-	set desc = "Type what you want to know about.  This will open the wiki in your web browser. Type nothing to go to the main page."
+	set desc = "Type what you want to know about.  This will open the /TG/ wiki in your web browser. Type nothing to go to the main page."
```
MrMelbert, line 4:
```suggestion
	set desc = "Type what you want to know about.  This will open the /tg/station wiki in your web browser. Type nothing to go to the main page." //NON-MODULE CHANGE
```

interface/interface.dm
```diff
@@ -20,7 +20,7 @@
 	set hidden = TRUE
 	var/forumurl = CONFIG_GET(string/forumurl)
 	if(forumurl)
-		if(tgui_alert(src, "This will open the forum in your browser. Are you sure?",, list("Yes","No"))!="Yes")
+		if(tgui_alert(src, "This will open the CL in your browser. Are you sure?",, list("Yes","No"))!="Yes") //NON-MODULE CHANGE
```
MrMelbert, line 23:
```suggestion
		if(tgui_alert(src, "This will open the changelog of server-specific changes in your browser. Are you sure?",, list("Yes","No"))!="Yes") //NON-MODULE CHANGE
```

PR #64, opened 2021-08-05
Post Round Pubby Fixes

Jolly-66:
Not sure if I got everything, but this is whats been changed:

HFR
Moved the room to be connected to the turbine. 
image

Xenobio
Turf was missing
image

Maint
Redid the wage cage a bit, spiffied it up. Hopefully I didn't break the disposals chute.
image

Misc
Added some meteor shielding because holy shit does Pubby not have it.
image
image

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -2298,10 +2298,6 @@
 	},
 /turf/open/floor/iron/dark,
 /area/security/execution/transfer)
-"aia" = (
-/obj/effect/turf_decal/bot/right,
-/turf/open/floor/iron,
-/area/engineering/atmos)
 "aid" = (
 /obj/structure/disposalpipe/segment{
 	dir = 4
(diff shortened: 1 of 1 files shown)


Jolly-66:
somethings to do and check
- intercoms range
- turf for drone bay and hfr
- misc decor
- pipes near wage cage

MrMelbert (changes requested):
Carpet still incorrect
image
Sink does not connect like this
image
This lattice is still stacked on a wall
image
Incorrect turfs used in this maintenance area, active turfs
image
This wall is still not fixed
image
big old intercom deadzone
image
Areas without an APC (most of them can be ignored, some are of note)
image
Many areas without an air alarm?  (Most can be ignored, some are of note)
image

MrMelbert:
image
image
doubled pipes
image
bad vent placement
image
image
unconnected vents and scrubbers
image
unconnected vents

MrMelbert:
image
only one HFR part?

MrMelbert:
image
telecomms tube still broken - going to works, returning does not - the tube dispenser in telecomms actual needs to be reversed

MrMelbert (changes requested):
Few more tiny nitpicks, I think after this it should be good to go

image
Camera placed on a door
image
still unconnected
image
few more areas without air alarms
image
This disposal does NOT lead to space like it should
image
missing plating under thsi window

PR #77, opened 2021-08-17
Subjecting myself to intense pain for minor benefits: Refactors loadouts to allow for more possibilities, and adds renaming items

MrMelbert:
- Entirely refactors loadouts from global lists to loadout datums. This allows for much much more potential with each loadout item and is slightly cleaner (?)

- You can now rename some loadout items. 
image

**Old loadout lists will be converted to new loadout lists automatically but I couldn't do the same to greyscale loadout lists, they've been wiped from this.**


MrMelbert:
Ah fuck I branched off the wrong branch

MrMelbert:
i guess i gotta cherry pick it

PR #80, opened 2021-08-20
Upstream merge to 8/20/21

MrMelbert:
It's weird... I can't help but feel like the amount of things that could break from this merge was immense but it compiled just fine and all the maps seems to work. 

I have a feeling I might need to explore more features to make sure nothing was deleted or broken.


Jolly-66 (approved):
CI looks good to me, if anything else needs to be done go for it

PR #86, opened 2021-08-29
Advanced Changelings, Neutered Changelings, and other refactoring

MrMelbert:
- Changelings now have support for the advanced antag panel, and can set their own objectives.
image
- Added another changeling antagonist datum, the Neutered Changeling, which is a very very weak version of a changeling. Much less chemicals, genetic points, can't absorb DNA, DNA sting, or transform. 
- Refactored a bit of advanced antag code.

- [ ] Add a way to make changelings to become a neutered ling if they were caught, or something like that
- [ ] Add other options for the ling panel
- [ ] Testing

### DIFF
--- a/code/modules/antagonists/changeling/changeling.dm
+++ b/code/modules/antagonists/changeling/changeling.dm
@@ -83,12 +83,12 @@
 	emporium_action.Grant(owner.current)
 
 /datum/antagonist/changeling/on_gain()
-	create_actions()
-	reset_powers()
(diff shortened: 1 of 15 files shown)


Jolly-66:
Please for the love of God nuke change sting
I'll forever be in your debt 

While we're here too, it would be a better idea to explore Ling as a super stealth antag, as that was some of my ideas for it's initial rework, go to see something of my ideas are making way here, intentional or not

PR #90, opened 2021-09-16
[MDB IGNORE] PubbyStation 2: The Retilening

CRITAWAKETS:
This PR changes Pubbystation to use new tiles instead of old crappy tiles, and changes the decals accordingly. I went for a design that tried to keep pretty much all the decals and such in the same places. Pretty much the only structural change is to cargo's front door being shifted one tile down in order to bring it in line with cargo, as making things more intuitive in cargo is good. It's a good department for new players, so that should be taken into consideration for map design.

This PR also adds in new decals, which are cut-outs of the new floor tiles. They arent used much in this map aside from the CE office, as they were made near the end of the development of this PR.

IMAGES:

There's too much to show, as almost every room of the map has been altered.

Please test this, i might have accidentally deleted something critical while making this, you never know with this sort of thing.

Fun fact: Because the decals are now grouped up, the map now has alot less decals, making it more performant!


Jolly-66 (changes requested):
If theres more, I'd like to see.

Personally not a fan of the chaplains office tiles, I think the old ones would work better here, or use the half tile ones. Them being smooth does not go over too good with me.

CRITAWAKETS:
https://i.imgur.com/z9L5ilN.png
https://i.imgur.com/UjPtL4X.png
https://i.imgur.com/c546o9V.png
https://i.imgur.com/6WcZzTp.png
https://i.imgur.com/GUh4hnB.png
https://i.imgur.com/ZdKgaPo.png
https://i.imgur.com/iLyJcS3.png

Im not embedding these because each and every single one of these is massive. This isnt even the whole map, as its literally the whole map that recieved changes.

jollystation_modules/code/game/objects/effects/decals/turfdecal/textured.dm
```diff
@@ -1,6 +1,6 @@
 /obj/effect/turf_decal/textured
 	name = "textured gray corner"
-	icon = 'jollystation_modules/icons/turf/textured_decals.dmi'
```
Jolly-66, line 3: ?

Jolly-66:
Discard that mobile review oops

PR #106, opened 2021-10-30
Adds a random event: Solar Flare

MrMelbert:
Adds a random event, titled "Solar Flare" to the game. 

This random event causes multiple "fireballs" to pop up everywhere around a department for the duration of the event. These fireballs are 2-4 radius diamonds of fire that deal moderate damage to mobs on the tile and of course set fire to anything around. 

- The event as uncommon, about as rare as a radiation storm.
- It can only occur once per shift without admin intervention.
- Once triggered, it will randomly begin sometime in the next 45 to 90 seconds.
- Once began, it will last somewhere between 90 and 150 seconds.
- There's a 30% chance the event will tell the crew where the solar flares are occurring. Otherwise, it will be unknown.
- A solar flare will trigger every 4 to 6 seconds.
- When a solar flare triggers, it will spawn a solar flare on 3 to 5 turfs in the target department. 
   - 88% of the flares will be standard 2 radius. They deal 60 damage if you're dead center, 30 damage adjacent, 15 on the rim.
   - 8% of the flares will be expanded 4 radius. They deal 90 damage if you're dead center, then 60, 30, 15. 
   - 4% of the flares will have 3 radius and come with an EMP. Heavy range encompasses only the dead center, light range encompasses the radius of the flare.

The goal of this PR is to add more environmental events to offset the lack of action. Gives medbay something to do and gives the crew something to worry about. Kinda like meteors but not roundending. 

image

### DIFF
--- a/jollystation.dme
+++ b/jollystation.dme
@@ -4049,6 +4049,7 @@
 #include "jollystation_modules\code\modules\clothing\under\jobs\security.dm"
 #include "jollystation_modules\code\modules\events\resource_drift.dm"
 #include "jollystation_modules\code\modules\events\resource_pods.dm"
+#include "jollystation_modules\code\modules\events\solar_flare.dm"
 #include "jollystation_modules\code\modules\food_and_drinks\recipes\drinks_recipes.dm"
 #include "jollystation_modules\code\modules\hydroponics\hydroponics.dm"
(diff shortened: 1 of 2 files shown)


Jolly-66:
You should def add the chance for them to cause an EMP strike in a small radius around it, just like a real one.

Doesn't even have to occur at all during its life time, but it'll also give engineers something to do if it causes enough damage.

PR #115, opened 2021-11-10
[modular] makes it so that unimportant items are deleted upon cryoing

Gofawful5:
there's definitely a better way of doing this.

deletes anything that doesn't have the indestructible tag
shouldn't conflict with anything in theory but uhhh y'know.

also makes it so borgs don't drop their mmis

image
image
image
image


Jolly-66 (approved):
I requested this,

Now keep your grimy fucking cursors away from looting cryo when someone sleeps.

PR #116, opened 2021-11-10
The worlds smallest fix: Cats now speak cat

Jolly-66:
no more human sfx on exclamation and ask says.

### DIFF
--- a/jollystation_modules/code/modules/mob/living/carbon/human/species_types/felinid.dm
+++ b/jollystation_modules/code/modules/mob/living/carbon/human/species_types/felinid.dm
@@ -4,6 +4,14 @@
 									'jollystation_modules/sound/voice/meow2.ogg' = 50,
 									'jollystation_modules/sound/voice/meow3.ogg' = 50)
 
+species_speech_sounds_ask = list('jollystation_modules/sound/voice/meow1.ogg' = 50, \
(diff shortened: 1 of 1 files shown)


jollystation_modules/code/modules/mob/living/carbon/human/species_types/felinid.dm
```diff
@@ -4,6 +4,14 @@
 									'jollystation_modules/sound/voice/meow2.ogg' = 50,
 									'jollystation_modules/sound/voice/meow3.ogg' = 50)
 
+species_speech_sounds_ask = list('jollystation_modules/sound/voice/meow1.ogg' = 50, \
```
MrMelbert, line 7: indentation is bad which is why it's complaining

jollystation_modules/code/modules/mob/living/carbon/human/species_types/felinid.dm
```diff
@@ -4,6 +4,14 @@
...
+									'jollystation_modules/sound/voice/meow2.ogg' = 50,
+									'jollystation_modules/sound/voice/meow3.ogg' = 50)
+
+species_speech_sounds_exclaim = list('jollystation_modules/sound/voice/meow1.ogg' = 50, \
```
MrMelbert, line 11: ditto

MrMelbert (changes requested):

PR #118, opened 2021-11-11
Makes small additions to research nodes easier to see

Wallemations:
Instead of completely overwriting code for the research nodes, lets you just add the IDs _you_ added, using a new variable list.

This way instead of _this_ for one single item addition
image

You have _this_ for one single item addition
image

Which makes it way easier to see what the actual addition is.


Wallemations:
I found out that using the same method to remove IDs works too! So I added that & some others

MrMelbert (approved):
good pr

PR #120, opened 2021-11-24
Cherry-picks strings fix to make master not fail

MrMelbert:
Cherry picks fixes to strings_replacement to make master not constantly fail.


Jolly-66:
does this mean i can unmute the repo and not have my inbox flooded every 20s

PR #124, opened 2021-12-03
Improves pubby engineering in many ways.

CRITAWAKETS:
I did a bunch of changes around the place, but i think the most telling one is this.

That's right, its a supermatter far enough away that the engineering rooms can actually be used.

There are also changes to decals around the place and changes to the HFR room.

Further improvements to engineering will happen at the same time as the science improvements, since there will be major structural changes, so stay tuned.


Jolly-66:
You forgot [MDB IGNORE]
I know it's small but eeeh

Jolly-66 (approved):
Not sure if tcomm sat will get touched by an explosion, but Pubby is an enigma and we love Pubby.

PR #129, opened 2021-12-08
Fixes Pubby Disposals

Jolly-66:
I didn't test this but I'm assuming these manifolds in sci maint were what was creating the loop. Hopefully its fixed.


Jolly-66:
i hate map prs

image

PR #136, opened 2022-01-14
Adds armor to basic MODsuits 

MrMelbert:
Adds armor to all the basic MODsuits. 

They're all ROUGHLY the same as the old versions of the hardsuits, adjusted down slightly to accommodate for all the added utility (and speed) that MODsuits bring. 

Additionally, syndicate MODsuits now have half their armor innate, and half their armor gained from activating their booster.


CRITAWAKETS:
Damn, this is pretty pog if i say so myself.

PR #139, opened 2022-01-23
Removes the old split Space Adaptation mutations

MrMelbert:
Removes hot, cold, and vacuum adaptation, and the recipe for space adaptation.

/tg/station nerfed space adaptation on their own so we don't need this anymore.


CRITAWAKETS:
i was actually planning on doing this today, smh

PR #149, opened 2022-02-27
Reploids now have appearance settings and IPC parts.

CRITAWAKETS:
Big PR to add in the first wave of appearance options for Reploids, allowing them to take on the form of humans with skintones or mutant colors, a unique limbset which is humans with a red triangle, and IPCs. This PR also allows reploids to take cat ears and IPC antennae on all limb types.

- [X] Fix the broken species trait change so the code actually compiles
- [ ] Find out what's causing reploid_antenna to become suddenly a null out of nowhere
- [ ] Clean up the code a bit
- [X] Better comments
- [X] Log the new non-module changes


CRITAWAKETS:
This has been impossible to get working for over a month and not even elder coders know what happened. Gonna wait for kapulimbs or some other refactor to get ported in before working on this again.

PR #150, opened 2022-03-04
Minor Lima changes

Wallemations:
## What I changed

- Removes a duplicate chair in virology
- Fixes the cargo shuttle's missing conveyor ID

Also:
- Adds bodybag item dispensers to multiple areas that don't already have bodybags but make sense to have them (I.e. genetics, virology, the morgue, robotics)
- Replaces a lot of the tables in medbay with glass tables
Map Screen
Glass tables in medbay look cleaner and nicer than normal tables imo.

### DIFF
--- a/_maps/map_files/LimaStation/LimaStation.dmm
+++ b/_maps/map_files/LimaStation/LimaStation.dmm
@@ -3063,7 +3063,6 @@
 /turf/open/floor/iron/dark,
 /area/science/breakroom)
 "boR" = (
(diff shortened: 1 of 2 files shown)


MrMelbert:
I was told the security order console was incorrect, can you fix that as well while you're here? :)

Wallemations:
> I was told the security order console was incorrect, can you fix that as well while you're here? :)

oh sure

PR #156, opened 2022-05-09
Overrides to rebalance techshells, mostly nerfs.

CRITAWAKETS:
So, tech shells have been nerfed. Here's the changelog.

Frag-12 no longer has the innate damage and stunning, the explosion already causes a LOT of damage.
Laser Scatter now does 66 damage over it's 6 pellets (11 per pellet) instead of 90 over it's 6 pellets (15 per pellet). This ends up being under 50 when wearing an armor vest!
Meteor slugs are now a true war crime of a projectile, no longer stunning their opponents and knocking them down for shorter times, but causing an incredible amount of pain.

### DIFF
--- a/maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -11,7 +11,7 @@
 //Overrides meteorslug to have no stun, lower knockdown but also causes a lot of PAIN
 /obj/projectile/bullet/shotgun_meteorslug
 	paralyze = 0
-	knockdown = 50
+	knockdown = 5 SECONDS
 
 /obj/projectile/bullet/shotgun_meteorslug/on_hit(atom/target, blocked = FALSE)
 	. = ..()
--- a/maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
+++ b/maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
@@ -16,6 +16,6 @@
 /obj/projectile/bullet/shotgun_meteorslug/on_hit(atom/target, blocked = FALSE)
 	. = ..()
 	if(iscarbon(target))
-		var/mob/living/carbon/M = target
-		if(M.pain_controller)
-			M.cause_pain(BODY_ZONES_ALL, 50) //OW MY BONES
+		var/mob/living/carbon/unfortunate_soul = target
+		if(unfortunate_soul.pain_controller)
+			unfortunate_soul.sharp_pain(BODY_ZONES_ALL, 50) //OW MY BONES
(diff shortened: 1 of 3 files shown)


maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
```diff
@@ -0,0 +1,21 @@
...
+/obj/projectile/bullet/shotgun_meteorslug
+	paralyze = 0
+	knockdown = 50
+
+/obj/projectile/bullet/shotgun_meteorslug/on_hit(atom/target, blocked = FALSE)
+	. = ..()
+	if(iscarbon(target))
+		var/mob/living/carbon/M = target
+		if(M.pain_controller)
+			M.cause_pain(BODY_ZONES_ALL, 50) //OW MY BONES
```
MrMelbert, line 21: Better variable name than `M`

MrMelbert (changes requested):

maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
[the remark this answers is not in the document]
CRITAWAKETS: Yeah sharp pain seems better, would certainly work better for security uses.

maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
```diff
@@ -0,0 +1,21 @@
...
+	paralyze = 0
+
+//New laser scatter projectile type, 11 damage for 6 pellets doing 66 damage in total
+/obj/projectile/beam/weak/laser_scatter
+	damage = 11
+
+//Overrides meteorslug to have no stun, lower knockdown but also causes a lot of PAIN
+/obj/projectile/bullet/shotgun_meteorslug
+	paralyze = 0
+	knockdown = 50
```
MrMelbert, line 14:
```suggestion
	knockdown = 5 SECONDS
```

maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
```diff
@@ -0,0 +1,21 @@
...
+
+/obj/projectile/bullet/shotgun_meteorslug/on_hit(atom/target, blocked = FALSE)
+	. = ..()
+	if(iscarbon(target))
+		var/mob/living/carbon/unfortunate_soul = target
+		if(unfortunate_soul.pain_controller)
+			unfortunate_soul.sharp_pain(BODY_ZONES_ALL, 50) //OW MY BONES
```
MrMelbert, line 21: iirc `sharp_pain` already checks for the existence of pain controller, so you can just apply it without concern (needing to sanity check)

MrMelbert (changes requested):

maplestation_modules/code/modules/projectiles/projectile/bullets/shotgun.dm
CRITAWAKETS: It doesn't, actually.

MrMelbert: image

PR #157, opened 2022-05-11
Adds in the new DMI editor plugin as a recommended VSCode plugin.

CRITAWAKETS:
Also removes an obsolete plugin for syntax highlighting that's essentially replaced by SpacemanDMM.

This is the DMI editor by AnturK, for clarification.


MrMelbert (changes requested):
```
JSON error in ./.vscode/extensions.json
Expecting ',' delimiter: line 9 column 3 (char 219)
```

PR #160, opened 2022-05-15
Completely destroys the balance of Pubbystation (adds a condimaster to the bar)

Wallemations:
# What have I wrought

Adds a condimaster to the bar of pubbystation so you don't need to go all the way to the coldroom to use it.
Also rearranges some other things in the bar to facilitate this change, and finally moves some boxes off the floor.

guh

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -2303,6 +2303,7 @@
 	dir = 4
 	},
 /obj/machinery/chem_master/condimaster,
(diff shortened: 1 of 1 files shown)


CRITAWAKETS:
powercrepe of the highest order

Wallemations:
image
bruh

Wallemations:
yea I have no clue if I should fix that integration test or not, it looks like it's an issue with resource caches, which I didn't touch

Wallemations:
I didn't even know our PRs _could_ turn stale

PR #161, opened 2022-05-29
TRAIT_NOBREATH no longer prevents spawning with lungs.

CRITAWAKETS:
Bug fix time! TRAIT_NOBREATH now no longer prevents species from spawning with lungs via the removal of some old, definitively deprecated code. Golems have been edited to have no lungs, as it makes sense for them to not have any.

Most notably, this fixes synths having no lungs.


CRITAWAKETS:
Gonna close since admittedly the rework is basically just a new thing so i don't wanna leave trailing files.

PR #163, opened 2022-06-03
Makes clockwork cult lore more consistent with server lore.

CRITAWAKETS:
Also more consistent with itself, as there were some issues.

Establishes some naming rules for Rat-Var. Rat-Var formally and Ratvar informally, Rat-var used mainly in adjectives. Was chosen for sounding more mechanical, like it's in sections done by each teeth of a cog.

In addition, a deconversion line has been changed to not have Rat-Var refer to their return in first person, as the clockwork cult is mostly trying to bring Ratty back to their side, rather than summon him.


CRITAWAKETS:
do not merge for now, headlorer insists on Rat'var.

PR #169, opened 2022-06-28
Adds in the ability to ricoshot projectiles off of coins.

CRITAWAKETS:
Projectiles can now ricoshot off of coins if you shoot at them, gaining a 1.5x increase to damage and a whopping 2x increase to speed. The coin flips afterwards, which adds in some flair alongside adding a small cooldown.

This allows some pretty risky and tricky maneuvers, but they could prove to be an effective source of damage. Of course, this is mainly just an easter egg for players of certain games (there's more than one game where shooting a coin increases damage).

If you shoot the coin with a projectile that's too powerful, you'll break the coin. No cheesing revolvers or laser cannons.


Wollywoger:
you might want to have diminishing returns if the projectile would surpass the threshold

PR #170, opened 2022-07-04
Adds Pride Pin to loadouts, adds reskin support to loadout, adds relaying support for loudout accessories

MrMelbert:
- Adds the Pride Pin to loadouts. 
- Adds reskin support to loadouts. If set, an item with possible reskins can be ... reskinned, out of the box, instead of needing to do it roundstart. 
- Adds layer adjusting support for loadout accessories. (Deciding whether or not an accessory renders above or below any worn suits.)

Kinda WIP while I decide whether or not I'd like to improve the code. It's currently a bit janky and hacky and could definitely be much better.


CRITAWAKETS:
YOOOOOOOOO

PR #171, opened 2022-07-12
Fixes the Phantom Pain + some upkeep

MrMelbert:
- Fixes the phantom pain bug with prosthetic limbs
- General upkeep for pain.


Wallemations:
> Fixes the phantom pain bug with prosthetic limbs

blasphemy

PR #172, opened 2022-07-21
Adds in story content, alongside an overlay for the character of Post.

CRITAWAKETS:
This PR adds in a folder for story content, alongside an overlay for the character of Post.

For the folder, all content is going to be seperated so it can be easily removed once rendered unused, as most of this content is going to be highly specific in usage.

Take for example, the Post Overlay, which only works with a felinid that has a very specific hairstyle.


CRITAWAKETS:
Both code review items are implemented and are tested to work.

PR #181, opened 2022-09-04
Mail now arrives on sundays

Wallemations:
I want my money and themed items


Wallemations:
Oh yea, forgot that
Unfortunately I won't be able to do that by the time the round starts

Think you or someone else could make a PR with that? If it's just a readme then it should be easy

PR #186, opened 2022-11-14
Adds CMO unique equipment to the garmant bag and locker + hair

MrMelbert:
I guess this technically doubles as event / story equipment AND general CMO stuff? At least it sounded like it.


Wallemations:
If you want this to be character-specific you could add it to the character creation wardrobe, name it something more specific so others know not to pick it

That's if you REALLY want this to only be for Vex though

Wallemations:
> Meh, I'd rather not put event/story content in the loadout wardrobe. If Drag wants it to be unique I can put it in an equipment beacon, if they want it to be generic it can stay in the CMO's locker. I don't mind either, but they're all written to be generic anyways.

Hadn't even thought about an equipment beacon
Yea that makes sense then

Draggeru:
> Meh, I'd rather not put event/story content in the loadout wardrobe. If Drag wants it to be unique I can put it in an equipment beacon, if they want it to be generic it can stay in the CMO's locker. I don't mind either, but they're all written to be generic anyways.

I don't particularly care where it's placed, depends on what the path of least resistance is. If a beacon is easier then do a beacon. If not then do the locker.

PR #189, opened 2022-11-22
Backports a remaster of pain that I forward ported

MrMelbert:
I updated Pain a good amount for another place and was pretty content with my changes
...which means that I wanted to bring it back here of course.

Lots of refactoring. Lots of non-modular changes, which is maybe pain for me later, who knows. 

Noticeable gameplay changes:
- Added two chems, Ibaltifen and Anurifen. These are good at healing brute and burn pain respectively.
- Medkits have Morphinemedipens
- Burn medkits have Anurifen medipens
- Brute medkits have Ibalitifen medipens
- Rebalanced a lot of the chemicals. Morphine is a lot better. 
- Rebalanced a few of the medipens. The miner medipen should be alot better. 
- Anesthetic is better handled. Being on Anesthetic gives an examine message, and - if in medial - a screen alert. 
   - Nitrous Oxide (Chem) now also gives anesthetic. 
   - Scrubbers in rooms which have anesthetic use will now start with scrubbers configured for it by default. 
- Surgery overall causes a lot more pain, unless the mob is on anesthetic or suitably numb.
- Shock expansion.
- A good amount of various bugfixing and small rebalances.


CRITAWAKETS:
Best part of testing on another codebase is that this is confirmed to work, as the earliest release had some **mood-destroying** bugs

PR #216, opened 2022-12-31
Gives Jellypeople lore

Wallemations:
Gives lore to one of the races in the list, which I thought appropriate since I'm (at the moment) the only one who plays as this race.

Obviously this is all very open to critique.


Wallemations:
Note: I need to add the word "permanent" to the general description, as it's probably not very rare to work with a temporary (mutation toxin) jellyperson

PR #218, opened 2023-01-02
Updates some old job stuff

MrMelbert:
- Deletes the old job config, updates the new job config. 
   - Scientist now has 3 spawn positions and 3 max positions by default
   - Ord tech now has 1 spawn position and 1 max position by default
   - Xenobiologist now has 2 spawn positions and 3 max positions by default
- The psychologist now points to two departments heads as implied by their supervisor text
- Adds some comments


CRITAWAKETS:
xenobiologist should have 2 max positions at most, even with 2 people it already gets crowded and chaotic in xenobio because of how it works

PR #237, opened 2023-02-06
Optimizes the tabletide smite

MrMelbert:
`for(thing in area)` is secretly a loop that does `for(thing in world) if(thing in area) ...`
This smite abused `in area`, which in turn meant it did a lot... a LOT of world iterations. Like 150+. 
Swaps it over to using the area contents system. Now it check ticks. 

Also added the more common bar area to teleport destinations. 

Two bugs I noticed with it that can be fixed later:
- Pacifist are immune to it.
- People with CQC will demolish every table.

### DIFF
--- a/maplestation_modules/code/modules/admin/smites/tabletide.dm
+++ b/maplestation_modules/code/modules/admin/smites/tabletide.dm
@@ -4,15 +4,19 @@
 
 /datum/smite/tabletide/effect(client/user, mob/living/target)
 	. = ..()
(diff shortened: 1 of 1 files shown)


CRITAWAKETS:
> `for(thing in area)` is secretly a loop that does `for(thing in world) if(thing in area) ...` This smite abused `in area`, which in turn meant it did a lot... a LOT of world iterations.

byond why must you do this

PR #252, opened 2023-03-13
Lesser flare customization

Ed640:
Allows flare to be customized in the spell book to a lesser cheaper version, lasts less than a minute and may fizzle out right after casting it.

image


nikothedude:
THIS REQUIRES MY FIX PR FIRST DONT MERGE IT WITHOUT IT

PR #255, opened 2023-03-20
Adds new height system for preferences

MrMelbert:
image
image

Adds new height system for preferences. 

This is a new way to give your character height variety without horribly mangling your sprite. 

Now it only does it a little, but it's barely noticeable. 

Hair might get cut off. Not sure if I got around to fixing that

The old system will persist for those who prefer it for the size factor. It's been renamed to "Character size".

### DIFF
--- a/code/datums/datum.dm
+++ b/code/datums/datum.dm
@@ -313,7 +313,7 @@
 
 /obj/item/update_filters()
 	. = ..()
-	update_item_action_buttons()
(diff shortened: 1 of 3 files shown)


nikothedude:
look at that little fucking guy holy fuck they have NO NECK

Wollywoger:
i'd like to call attention to the branch name

PR #271, opened 2023-05-03
PUBBY TIME NEVER ENDS

CRITAWAKETS:
That's right, Pubbystation is back. With a shitload of bug fixes too, the map's likely never been better!

Okay, I lied, there's probably a bunch of access IDs missing off of stuff because of the access reworks, BUT IT'LL BE FINE!!

Ordnance has also recieved a small overhaul to add a freezer chamber and tank compressor, enjoy.


CRITAWAKETS:
every check passes, a demon is unleashed

PR #272, opened 2023-05-05
blush noise when you *blush, as mentioned by host.

Andrymeda:
on the request of Crit, adds in a lil blush noise from another certain server ( pensive )

image
 

https://user-images.githubusercontent.com/109443978/236370025-5a75e929-5b4d-42b6-a4f2-3e3839bac5e6.mp4

### DIFF
--- a/maplestation_modules/code/modules/mob/living/emote.dm
+++ b/maplestation_modules/code/modules/mob/living/emote.dm
@@ -16,5 +16,5 @@
 	stat_allowed = SOFT_CRIT
 
 /datum/emote/living/blush/run_emote(mob/user, params, type_override, intentional)
-	..()
-	playsound(user,'maplestation_modules/sound/emote/blush.ogg',80)
+	. = ..()
+	playsound(user, 'maplestation_modules/sound/emote/blush.ogg', 80)


maplestation_modules/code/modules/mob/living/emote.dm
```diff
@@ -14,3 +14,11 @@
...
 	message = "exhales sharply."
 	stat_allowed = SOFT_CRIT
+
+/datum/emote/living/blush/run_emote(mob/user, params, type_override, intentional)
+	. = ..()
+	if(. && ishuman(user)) // Give them a visual blush effect if they're human
+		var/mob/living/carbon/human/human_user = user
+		ADD_TRAIT(human_user, TRAIT_BLUSHING, "[type]")
+		playsound(user,'maplestation_modules/sound/emote/blush.ogg',80)
+		human_user.update_body_parts()
```
CRITAWAKETS, line 24:
```suggestion
/datum/emote/living/blush/run_emote(mob/user, params, type_override, intentional)
	..()
	playsound(user,'maplestation_modules/sound/emote/blush.ogg',80)
```

CRITAWAKETS (changes requested):
you dont need to duplicate the emote's code, just do it like this! no human check is needed aswell since it's only to set an overlay on humans which would otherwise not fit with say, mob sprites.

maplestation_modules/code/modules/mob/living/emote.dm
CRITAWAKETS: you'll want to edit this in manually because i can't do a multi-line request change in github

CRITAWAKETS:
it failed on an icebox powernet, not the code's fault lol

PR #276, opened 2023-05-15
Greatly enhances the surgical abilities of Limastation

Wallemations:
image

Adds another door from the main medical area so you don't need to drag someone through the hall to get to the operating room
Hopefully this makes people actually use it more 💀


token-map:
SPEEDMERGE IMMEDIATELY

PR #283, opened 2023-06-10
[MDB Ignore] [IDB Ignore] [Still have neither] 6 months of Upstream 

MrMelbert:
I re-wrote the merge driver script from bash to python in an effort to make it a bit faster. 
Before, it really choked up on the "delete unticked files" part. 
Now, it chokes up on "resolve merge conflicts" part. 
Huge success.

I refactored some miscellaneous code, like speech / radio sounds. Fixed a bug with that.  

Moved Hurtful Disposals to a better folder. Though I want to move it to an even better folder later. 

Also, as a consequence of the aforementioned script rewrite I may have accidentally wiped some changes from some files. 
We went from 143 non-module comments to 115. Some of those were intended. Some of those were not. 

- [x] Find the 30-odd mission non-module comments
- [ ] Records are inaccessible due to TGUI records consoles
- [x] Probably some others


CRITAWAKETS:
im betting a few of those missing non-module changes are going to be on our numerous attempts to hack speciescode to try and get things working, cant wait for skrell tentacles to disappear again

PR #287, opened 2023-06-21
Cresssprites pt. 1 (clothes and stamp)

Higgin:
Adds new items, sprites, and spawners for thedragmeme's cresstation and character.

image
image
image
image

This includes a purple resprite of three parts of the noble_equipment set (with a new spawner for the specific pieces that were redone) as well as the full 'winter princess' set of gear intended for the noble ambassador role on the eventual map PR. (The 'ambassador' stamp is also included in this PR.)

This PR includes a change to make our toggles.dm behave like /tg/'s. One unintended side effect of this (see the last picture) is that until the upstream PR splitting up coats, hoods, and allowing winter coats to be zipped is merged, https://github.com/tgstation/tgstation/pull/74886, winter coats will show their hooded/fully-zipped icon-state on the outerwear slot when the hood is up. It's minor, but it should resolve with an upstream merge as it's a direct copy of the /tg/ file with exception to two procs we also don't use yet (for updating clothing buttons.) I adopted this behavior because it was also the cleanest way to get the sprites as formatted to work and to allow the cloak to be worn as either an exosuit or neck-slot item.

Please let me know what if anything I might need to do to this, or if that's alright - God knows I could squash the commits, but I don't want to lose the history in case I need to change something around the toggles.


Draggeru:
I always enjoy it when people indulge in my laziness

Draggeru:
Assuming Crit added the stamp #422 should be able to close this

PR #291, opened 2023-06-28
Adds autopsy scanners to the morgue and robotics

siliconOpossum:
Adds a couple of autopsy scanners roundstart to make dissections less of a pain when there isn't a coroner around

Morgue
image
Robotics
image


CRITAWAKETS:
honestly wouldn't be too needed if the upstream didnt make autopsy scanners weirdly either require a crate from cargo or buying a _premium item_, alongside having zero skeleton access on the secure morgue for some reason

PR #299, opened 2023-07-03
Pain tweaks

MrMelbert:
Lots of pain tweaks. 

- Low levels of pain no longer slam you with a effects immediately. Pain side effects now scale directly with amount of pain, and it's...relatively linear. 
   - Determination stops all pain effects. It did before but just re-iterating this.

- There is no longer a hard cap to pain per bodypart. It has been replaced with a soft cap. Limb pain per bodypart can exceed the soft cap, but pain above the soft cap is healed faster. 
   - Why? I noticed it was extremely easy for someone hitting the exact same bodypart to cap the limb's pain and then... all further attacks just did nothing. This meant if your leg was completely devastated, your average pain sit at like sub 20%, which caused pretty much no side effects. 

- Shock is no longer RNG induced, and is now built up to. The value required is based on your current health.
   - Why? Again, pretty much no one entered shock because no one had more than 20% pain. And even if you did collect an absurdly high amount of pain, you can shrug it off with good RNG. Likewise if you just dipped in high pain for a second, you could unluck into shock. 
   - Now the system is slightly more interactive. Upon entering thresholds your shock will start to build up slowly, but stuff like epinephrine can negate or slow down the rate of increase. 
   - If you're super banged up and in a lot of pain, then yeah you're gonna enter shock. Logically shock should serve as a replacement / augment to soft crit. 
   - Determination stops shock. 

- General refactoring, optimization, and bugfixing.

- Forced screams no longer ... force, if you can't feel pain at the moment. 

- Dropping an item from pain now has a unique message if your arm is robotic instead.


CRITAWAKETS:
getting closer and closer to rimmed

also thank goodness for the tram no longer putting people into shock via RNG

maplestation_modules/code/datums/pain/pain.dm
```diff
@@ -492,31 +518,121 @@
...
+
 			if(70 to INFINITY)
-				high_pain_effects(seconds_per_tick)
+				shock_buildup += 3
+				if(SPT_PROB(2, seconds_per_tick))
+					to_chat(parent, span_userdanger(pick("Stop the pain!", "Everything hurts!")))
+
+		// If shock buildup exceeds our health + 30 ticks then well, we enter shock
+		// This means at 100 health you can be in moderate pain for 130 ticks / 260 seconds / ~4 minutes before falling into shock
+		if(shock_buildup >= (parent.health + 30) \
```
CRITAWAKETS, line 571: shouldn't shock be done via a proc rather than in process

MrMelbert: Objectively speaking this is faster :tm: because it removes some proc-call overhead from rather hot code 
But it doesn't really matter all that much

PR #300, opened 2023-07-06
Adds early version of Limb Manager, removes prothestic limb quirks, adds arm-implanted lighter

MrMelbert:
- Adds an early version of the Limb Manager.
   - This replaces prosthetic quirks as a way of adding cybernetics to your character at game start. 
   - There's still some stuff I'd like to do to it, like set it into the preference menu itself (also with the loadout manaager), add icons rather than checkboxes, and clean it up a bit, but it's serviceable now. 

image

- Adds an arm implant, the implanted lighter, to the Limb Manager. 
   - With the snap of a finger, you can light something on fire. So cool. 

image


MrMelbert:
Merging this in a functional-but-incomplete state so that people can throw in ideas / new organs / etc if they have any

CRITAWAKETS:
didnt expect it to be ready that fast, goddamn

PR #315, opened 2023-07-24
Adds a lil overlay to people with temp flavor text

nikothedude:
image

On skyrat, this lil image tripled the usage rate of temp flavor text and made people see it way more


maplestation_modules/code/modules/flavor_text/pose_component.dm
```diff
@@ -18,6 +25,10 @@
 		SIGNAL_REMOVETRAIT(TRAIT_INCAPACITATED),
 	), PROC_REF(on_incapacitated))
 
+	var/mob/living/living_parent = parent
+	living_parent.add_overlay(list(GLOB.pose_overlay))
+	living_parent.update_overlays()
```
MrMelbert, line 30: - `add_overlay` does not require you pass it a list, so the wrapping is un-necessary
- Calling `update_overlays` also does nothing, it only actually updates overlays if called via `update_appearance`. (`update_appearance(UPDATE_OVERLAYS)`

I think it'd be less liable to break if component just registered `COMSIG_ATOM_UPDATE_OVERLAYS` and added 
the overlay via that

maplestation_modules/code/modules/flavor_text/pose_component.dm
```diff
@@ -1,3 +1,10 @@
+GLOBAL_VAR_INIT(pose_overlay, generate_pose_overlay())
+
+/proc/generate_pose_overlay()
+	var/mutable_appearance/temporary_flavor_text_indicator = mutable_appearance('maplestation_modules/icons/misc/temporary_flavor_text_indicator.dmi', "flavor", FLY_LAYER)
+	temporary_flavor_text_indicator.appearance_flags = APPEARANCE_UI_IGNORE_ALPHA | KEEP_APART
+	return temporary_flavor_text_indicator
```
MrMelbert, line 6: Also this can just be a static var on the pose component itself no reason to make it global

MrMelbert (changes requested):

maplestation_modules/code/modules/flavor_text/pose_component.dm
nikothedude: ill change it to work like that when i get home

MrMelbert (approved):

PR #320, opened 2023-07-25
Converts loadouts, limbs, and languages to preference middleware

MrMelbert:
They in tabs now. 

image


RimiNosha (commented):
On my end, I've been able to get rid of just about all the `name` fields on the item datums by using `style={{ 'text-transform': 'capitalize' }}` on `_LoadoutManger.tsx:147` along with using initial item names. It so far plays well with everything, including the weirder names like Nar'Sie.

Hopefully the removal of the need to type out a million item names will make adding new items less tedious.

You'll need to edit the category datum to check just the item_type instead and not throw an angry message, though, but no further edits are required.

maplestation_modules/code/modules/loadouts/loadout_ui/loadout_manager.dm
```diff
@@ -68,113 +73,86 @@
...
-	var/save_on_close = FALSE
 
-/datum/loadout_manager/New(user)
-	owner = CLIENT_FROM_VAR(user)
-	owner.open_loadout_ui = src
-	loadout = owner.prefs.read_preference(/datum/preference/loadout)
+/datum/preference_middleware/loadout/New(datum/preferences)
+	. = ..()
+
+	loadout = src.preferences.read_preference(/datum/preference/loadout)
```
RimiNosha, line 82: This is called before preferences are properly initialized, also is called with null preferences due to how static caching works.

I'd also just recommend getting rid of the `loadout` var, and instead read from prefs when you need the list inside procs. It's cached already anyways, and you're already passing prefs via `manager` where you'd need this.

Or if you really want to be as cool on the cycles as possible, attach this to the preferences datum, and handle it from inside there. Doing all this inside middleware won't work without monmod edits.

MrMelbert:
> Hopefully the removal of the need to type out a million item names will make adding new items less tedious.

With hindsight the name should default to the initial item name if left unset. Makes it less tedious (as you say) but still allows for humans to edit certain item names if they want to be more specific.

maplestation_modules/code/modules/loadouts/loadout_ui/loadout_manager.dm
MrMelbert: Ah yes that does explain the runtimes I saw but had not bother to fix yet. Good shout.

maplestation_modules/code/modules/client/preferences/languages.dm
```diff
@@ -82,122 +77,94 @@
...
 /datum/language/piratespeak
 	available_as_pref = TRUE
 
 /datum/language/yangyu
 	available_as_pref = TRUE
+	banned_from_species = /datum/species/ornithid
+
+/datum/language/shadowtongue
+	available_as_pref = TRUE
+
```
Jolly-66, line 119:
```suggestion

```

You have an accidental extra space here.

MrMelbert:
I think this is ready, if not, we'll fix it in post

PR #325, opened 2023-08-07
adds green tea as an upgrade reagent to soda machine

Wallemations:
I wouldn't be surprised if trying to use a |= on a null var breaks it but since it's originally a list in the first place I think I'll just go with it

I doubt we're gonna get tea leaves for green tea so I figured making it an upgrade would be good


CRITAWAKETS:
as the creator of green tea i support this, we aint /tg/ so i'm fine with a printable reagent after upgrades that's still a level below what chemistry can make

Wallemations:
Ok cool so I can't just do |= according to the tests

MrMelbert:
> I wouldn't be surprised if trying to use a |= on a null var breaks it

This behavior works fine with `+=` but not `|=` because byond. 

You can always use `LAZYOR` though.

PR #351, opened 2023-09-27
Adds "Stowaway" job

MrMelbert:
Adds the "Stowaway" job, subconsciously inspired by the Goonstation trait of the same name (where I ended up making this entirely independently and then remembered the Goon trait exists).

You get: 
- No bank account
- No headset
- No PDA
- No proper ID card
   - You're given an ID card which has basic maintenance access, but does not have your name
- A toolbox with some tools
- A station bounced radio
- A flashlight
- Some pocket change

Additionally
- You have no entry on the crew manifest, no security record, and no medical record
- You don't receive mail
- You cannot latejoin as this job
- You spawn somewhere random in maintenance

**But why?:** 

I thought it'd be a fun job for one-off gimmicks.

I don't expect anyone to ""main"" Stowaway, but if they want a way to introduce a new character, or have a character they rarely play, they can cut down on some admin setup by just throwing themselves in maintenance and hiding until they get found. 

Or it gives some shenanigans for security to get up to as a stowaway is technically not here legally. Of course you're not an antag


CRITAWAKETS:
Would be interesting if there were a slight element of randomness to it. Presumably most stowaways were there from a previous shift or the station construction crew/reclamation crew, so they'd have some kind of story behind that. It's an easy way to generate a fun little story for your character if you have no idea what to do.

code/modules/modular_computers/file_system/programs/jobmanagement.dm
```diff
@@ -28,6 +28,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 		JOB_ASSISTANT,
 		"Bridge Officer", //NON-MODULE CHANGE // MELBERT MODULARIZE ME
 		"Asset Protection", // NON-MODULE CHANGE
+		"Stowaway",
```
CRITAWAKETS, line 31:
```suggestion
		"Stowaway", // NON-MODULE CHANGE
```

MrMelbert: This list is killed next update anyways, so putting comments by it is pointless

MrMelbert:
> Would be interesting if there were a slight element of randomness to it. 

I kind of intended it to be a "you fill in the blanks" role, but I could generate some fake memories for people who just wanna get something random.

maplestation_modules/code/modules/jobs/job_types/stowaway.dm
```diff
@@ -0,0 +1,207 @@
...
+			backstory_equipment = "A syndicate turtleneck and mask, and some insulated combat gloves."
+
+		if(4)
+			var/old_boss = pick_list(COMPANY_FILE, "good_companies")
+			backstory_gist = "You used to work for [old_boss], but you were fired on the job, and have ended up marooned on this station. \
+				You're not sure what to do now, but you're sure you'll think of something."
+			backstory_suggested_goal = "Give up your old life and start anew - maybe on the station itself, or attempt to reconcile with [old_boss]."
+
+		if(5)
+			var/datum/job/old_job = pick(SSjob.joinable_occupations)
```
CRITAWAKETS, line 99: didnt notice this, but wouldnt this mean you can get a head of staff
would be funny if you ended up getting fucking _captain_

PR #353, opened 2023-10-04
Adds sprinting (inspired from Citadel's system), makes walking default

MrMelbert:
### Note

I recommend the walk delay be changed to 2.5 and the run delay, 1.5. 
Just as a suggestion, as those are the values I tested it with, of course you can make it whatever you want. 

### Details

- Humans default to walk move intent. 
- Being on run intent drains a sprint bar, shown on your move intent selector. 
   - You can run for 50 tiles (by default) without penalty (indicated by the sprint bar). 
   - Running for longer than 50 tiles incurs a stamina cost for every tile sprinted. 
   - You regenerate 0.5 tiles of sprint per second (Or 1 per life tick) while not sprinting. 
   - Being stamcrit fully drains your sprint bar. 
   - Stamina regenerating chemicals (such as stimulants) regenerate your sprint bar.
- Walking no longer masks your footsteps. 
   - It's a little goofy if no one gave off any footsteps at all unless sprinting. 
   - This is kinda sad to me because I use it often, maybe it can return as a "stealth stance" or something?

### Other

Sound effects ported from Citadel Station. 

### Todo 

- [ ] Being forced to the ground still consumes sprint while resting does not.


Wallemations:
Copy-pasting for posterity;

> I can't pass any creatures while in walking intent, I think it's an interaction with pushover or something
> 
> Also, shouldn't the stamina bar refill while you're standing still even if you're in run mode? It feels weird to have to swap out of the mode even if I'm standing still

PR #354, opened 2023-10-06
Kidnaps bottle sprites from upstream

Wallemations:
image

tin

I want my fuckin overlays to not look weird


Wallemations:
Ohhhh right they're both in the temp folder, gotta fix the .dme

PR #356, opened 2023-10-09
Being defibbed abates shock and heals some pain 

MrMelbert:
Idea came to me in the last round. The idea behind this is that if someone's in turbo shock, it helps expedite getting them back into the round in a stable state. 

Is it realistic? I dunno, but it's worth a try.


CRITAWAKETS:
how this idea came to melbert: i got thermonuclear bombed in the face and ended up being in shock for forever

PR #362, opened 2023-10-30
Adds liver and lung autosurgeons to cargo

MrMelbert:
- Adds Liver and *LUNG Emergency Autosurgeons to cargo.
   - These cost 2000 credits, but of course can be freely ordered with departmental orders. 
   - Requires surgery access. 
   - Cannot be self-used instantly, like other auto-surgeons. 
      - (They CAN be self-used, just not instantly. It requires a do_after as if used on another person.)
   - The only accept their corresponding organ types. 
   - These autosurgeons only work on mobs with very damaged or otherwise missing organs. 

Replacing the livers and lungs of people who repeatedly drink themselves is fun and all for the medical staff, but after the nth patient on the same round it maybe gets a little old. 

So, if an MD is starting to get impatient, they can pop in an order for one of these to make their life going forward easier. 
Someone comes in with a failing liver, just pop a new cybernetic liver in the autosurgeon and let it do the work.

### DIFF
--- a/maplestation.dme
+++ b/maplestation.dme
@@ -5683,6 +5683,7 @@
 #include "maplestation_modules\code\modules\research\techweb\_research.dm"
 #include "maplestation_modules\code\modules\research\techweb\all_nodes.dm"
 #include "maplestation_modules\code\modules\surgery\organs\augments_arms.dm"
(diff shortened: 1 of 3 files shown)


CRITAWAKETS:
> * Adds Liver and Ling Emergency Autosurgeons to cargo.

liver and WHAT NOW

PR #366, opened 2023-11-01
Adds `R_BAN`, `R_DEBUG`, or `R_SERVER` requirement to seeing ahelp ticket popups / tab

MrMelbert:
By request: Admins require one of `R_BAN` (banning permissions) `R_DEBUG` (debug permissions) or `R_SERVER` (server restart permissions) to see ahelps. 

Also: This is very much not foolproof, as long as someone knows how VV works they you can unwrap the ahelp tracker pretty easily. However it should serve for our purposes (and is very easily caught). 

Also also: This also half affects interviews because the panels are intertwined, but I couldn't bring myself to care about it since we don't do in game interviews. In fact I imagine they're disabled.

Also also also: this is difficult to test, so should be test-merged on a pre-round or something.

### DIFF
--- a/code/controllers/subsystem/statpanel.dm
+++ b/code/controllers/subsystem/statpanel.dm
@@ -115,6 +115,8 @@ SUBSYSTEM_DEF(statpanels)
 	target.stat_panel.send_message("update_mc", list("mc_data" = mc_data, "coord_entry" = coord_entry))
 
 /datum/controller/subsystem/statpanels/proc/set_tickets_tab(client/target)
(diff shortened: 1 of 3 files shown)


CRITAWAKETS:
Why debug permission specifically? It's given out to event runners since it's the only way to do things like manually spawning in cargo crates or map templates.

MrMelbert:
> Why is debug permission in the list specifically? It's given out to event runners since it's the only way to do things like manually spawning in cargo crates or map templates.

I left it in for when someone (me) needs to respond to ahelps involving bugs. 
When I gave it a quick glance earlier I didn't see a lot behind `R_DEBUG` that would make it a must-have for running events. 
You can always change any verbs which require debug to not require it.

PR #367, opened 2023-11-02
Reworks CPR

MrMelbert:
- [x] Needs a bit of testing first. 
- [x] Also medical staff need CPR skillchips(?). 

### What?

- CPR is no longer entirely rescue breaths: 
   - When beginning CPR, you engage in chest compressions.
      -  Untrained members of the crew doing CPR have the potential to cause harm with these chest compressions, if they begin panicking. 
         - Members of medical start CPR trained.
         - Also, there's a neutral quirk that also grants you CPR training. 
            - (I might make it a 1 point quirk. Not sure. Seems limiting.) 
      - Chest compressions to not heal oxyloss, but prevent heart attacks from further harming patients. 
      - There's also a 1% chance to restart the heart of a patient. Do you feel lucky?
   - After a set number of compressions, you deliver a rescue breath.
      - These rescue breaths are what heal oxyloss. 
- CPR "Panic state", now begins after a full cycle of CPR. 
- I rewrote heart attacks to accommodate this. They should overall act more consistent now, as they have been signalized / event based rather than work off life processing. 

### Why?

- Adds a bit of depth to the medical system, particularly CPR. Current CPR is rather unrealistic. 
- Given the existence of shock, having methods to stave off heart attacks in the field is good.


CRITAWAKETS:
Renember to port the heart attack rewrite upstream. It's free GBP.

Draggeru:
We should un-nerf belt defibs, since those are the closest things we have to an AED counterpart.

Or you know, make an actual AED counterpart.

MrMelbert:
> We should un-nerf belt defibs, since those are the closest things we have to an AED counterpart.

Define un-nerf belt defibs

> Or you know, make an actual AED counterpart.

Now we're talking

PR #368, opened 2023-11-02
Compact defibs fit on labcoats and similar medical gear

MrMelbert:
Compact defibrillators fit on suit storage slot for labcoats and similar medical outerwear (surgical apron, winter coat). 

It also fits on Medical MODsuits, but that one I'm not sure is necessary since the defibrillator MOD exists and is cooler. 

image


CRITAWAKETS:
Do the worn sprites appear correctly or is it just a pink and black blob?

PR #369, opened 2023-11-03
Re-adds vampire bat form, disallows vampires from entering other departments without permission

MrMelbert:
Happy Halloween guys, am I late?

- Re-adds Bat Form
- Vampires are not allowed to enter departments that they do not have permission to. 
   - They can obtain permission by asking (verbally) to enter. If a member of that department responds allowing it, you may enter as normal (assuming your ID card has valid access, as well). 

image

- [ ] Add some more trigger words to the regex
- [x] Ensure ventcrawling works


MrMelbert:
I'll add more triggers later (clueless)

PR #370, opened 2023-11-05
Adds in the Persuade spell.

CRITAWAKETS:
Lore and spell icon not done yet, so it's a draft for now.

This PR adds in the "Persuade" spell, a very customizable spell that allows psychic persuasion of any target. It can be configured to require an emote on cast, and/or only work if the target has low sanity. Or neither. It comes with easy to edit variables and admin logging for, well, the admins.

image
(note: this doesn't give forced speech to the target, I was just doing it for the reference.)


CRITAWAKETS:
People don't really like this spell, I don't blame them honestly. I might end up coming back to this whenever I end up working on evil psychic shenanigans.

PR #373, opened 2023-11-09
Double Decker Lima AI satellite

MrMelbert:
Lima's AI sat / telecomms is now doubled up on top of each other rather than awkwardly beside one another

This has a unique gimmick in that the AI core is now linked to the telecomms server room, via catwalks. Cold AI. Very cool

image

image


notfrying1pans:
does the dog come with it

PR #375, opened 2023-11-13
CPR tweaks

MrMelbert:
- Lets people CPR on people who don't actually need it (are just suffocating). 
- Compressions help a little with oxyloss to accommodate.


githubuser4141:
Question, if someone's suffocating - how does this actually help? CPR isn't healthy to do on someone and it's generally unrecommended to do on someone who's breathing and has a pulse. If someone's suffocating in a room because there's no air then how the heckarinos does breaking  their ribs help? It doesn't.

MrMelbert:
It does not, but there is some degree of gameplay abstraction here. Spacemen are incredibly conditioned to CPR people who are suffocating.

PR #377, opened 2023-11-14
Adds loadout slots / presets (on a per-character basis)

MrMelbert:
Players can now have up to 5 loadout presets saved per character.

image

Other changes:
- Optimized height filter in preference menu, making it a lot less laggy for the server
- Other filters now show in the prefs menu, because I re-added keep together to the dummies. 
- Cherry-picks appearances being visible in VV (for debugging)

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -376,7 +376,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	body = new
 
 	// Without this, it doesn't show up in the menu
-	body.appearance_flags &= ~KEEP_TOGETHER
+	// body.appearance_flags &= ~KEEP_TOGETHER // NON-MODULE CHANGE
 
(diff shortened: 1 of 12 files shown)


code/modules/client/preferences.dm
```diff
@@ -376,7 +376,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	body = new
 
 	// Without this, it doesn't show up in the menu
-	body.appearance_flags &= ~KEEP_TOGETHER
+	// body.appearance_flags &= ~KEEP_TOGETHER // NON-MODULE CHANGE
```
MrMelbert, line 379: I removed this code to get filters to show up

On one hand: The dummies show up in the menu, meaning the comment may be outdated
But on the other hand: There is an issue where only the loadout dummy disappears when the player joins the game, so ???

PR #378, opened 2023-11-20
Adds Blood Volume and ""Heart Rate"" to operating consoles

MrMelbert:
Operating consoles now reports patient blood volume and ""Heart rate"". 

We don't have real heart rate, so this number is faked, based on some conditions about the patient. _(One day it will not be faked, though.)_ The primary intent of this is to tell you when their heart is stopped and they require CPR or defibrillator, but it can also show you stuff like "are they in crit" if their heart rate is very high.


CRITAWAKETS:
the first of organmed begins

PR #385, opened 2023-12-11
Adds Tiziran Blood Wine but doesn't break the .dme in the process

Bytube:
whoops.

Files should work now without conflicts.


maplestation_modules/code/modules/reagents/reagent_containers/cups/glassbottle.dm
MrMelbert, line 1: I don't think this file is ticked.

maplestation_modules/code/modules/reagents/reagent_containers/cups/glassbottle.dm
```diff
@@ -0,0 +1,11 @@
+#define BOTTLE_KNOCKDOWN_DEFAULT_DURATION (1.3 SECONDS)
+
+/obj/item/reagent_containers/cup/glass/bottle/blood_wine
+	name = "bottle of lizard wine"
```
MrMelbert, line 4: `"bottle of lizard wine"` is used for, well, lizard wine. So you may wish to change this. Unless it's intentional it shares a name.

maplestation_modules/code/modules/reagents/reagent_containers/cups/glassbottle.dm
```diff
@@ -0,0 +1,11 @@
+#define BOTTLE_KNOCKDOWN_DEFAULT_DURATION (1.3 SECONDS)
```
MrMelbert, line 1: You don't need to copy this define

maplestation_modules/code/modules/reagents/reagent_containers/cups/glassbottle.dm
Bytube: how would I go about ticking the file?

MrMelbert: If you use VScode, hover over this in the bottom right and click it. (It should say "Unticked", and change on click). 
image

PR #405, opened 2024-01-26
[MDB Ignore] [IDB Ignore] [MDB IGNORE] [IDB IGNORE] Upstream Merge, filter created digitigrade clothing sprites, some other stuff

MrMelbert:
Upstream merges to like, January 14th (the branch name is a lie) 

Maybe the last one? Who knows. Playing it fast and loose from here on. 

Big things include Bitrunners! Very cool wow
More heretic paths, doesn't really apply but can for antag gimmicks. 
Food causes disease if left on the floor, probably need to tweak that. 
Security mech. 
New quirks?

See here: https://github.com/tgstation/tgstation/pulls?page=1&q=is%3Apr+is%3Amerged+closed%3A%3E%3D2023-06-01+sort%3Acomments-desc

In other news: 

I deleted all the outdated digitigrade sprites and added a system to use displacement maps for digitigrade clothing sprites. It's not perfect, the filters could use some fine tuning, but since I'm the only one who uses them it's my cross to bear.


CRITAWAKETS:
you forgot the biggest and baddest change out there, this also switches from 514 to 515, which is neat and has some implications for server performance.

PR #413, opened 2024-01-31
Buffing leylines

Ed640:
Increases base mana recharge.
Leyline strength picked will be average or higher, so will always be at base mana recharge rate or higher.


MrMelbert:
We see for now

PR #424, opened 2024-02-08
Tweaks event frequency

MrMelbert:
Lower end from 2.5 minutes -> 5 minutes
Upper end from 7 minutes -> 15 minutes


notfrying1pans:
mtereors :(

PR #426, opened 2024-02-11
Fixes Statues emitting solid lights (#81398)

Ed640:
Ports Statue light fix from https://github.com/tgstation/tgstation/pull/81398


Ed640:
I am not good with github, was i supposed to put [mirror] in the title or anything like that?

PR #428, opened 2024-02-12
Overrides valentines to get rid of the romantics.

CRITAWAKETS:
This PR just overrides Valentine's Day to not give the dates, which suck on a server with statics. Best do it now before we forget about it a year later.

Also removes the valentines cards. No reason why other than they suck and I hate them.


MrMelbert:
We're gonna change it to be an opt in thing when it triggers instead

CRITAWAKETS:
https://github.com/tgstation/tgstation/pull/81499
This'll just get ported (albeit with a few small tweaks) when ready.

PR #441, opened 2024-02-29
More Synth Fixes

MrMelbert:
- Fixes undisguised synths spamming messages on taking damage


MrMelbert:
Actually doing the rest of this in the blood PR since they're related now

PR #449, opened 2024-03-07
Fixes round and station time.

CRITAWAKETS:
This PR fixes round and station time being incredibly broken on any machine that has a timezone that isn't "0".

Non-modular changes that aren't marked ahead, but it's likely upstream is going to either pretty much port this or refactor the code outright.

Also station time doesn't exactly have feature parity with how it's supposed to work, but only really manifests in some slightly different hour offsets. The original way it worked _does not work in 515_, as negative values no longer work in time2text.


notfrying1pans:
18:59:59

PR #459, opened 2024-03-18
You walk faster with canes on missing limbs

MrMelbert:
Holding a cane reduces the movespeed penalty for missing limbs by 50%. Per cane, so you can have two canes for 75%. If you want

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -2289,14 +2289,18 @@ GLOBAL_LIST_EMPTY(fire_appearances)
 			if(!usable_hands)
 				ADD_TRAIT(src, TRAIT_IMMOBILIZED, LACKING_LOCOMOTION_APPENDAGES_TRAIT)
 
+	update_limbless_movespeed_mod()
+
+/mob/living/proc/update_limbless_movespeed_mod()
 	if(usable_legs < default_num_legs)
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/living.dm
```diff
@@ -2289,15 +2289,19 @@ GLOBAL_LIST_EMPTY(fire_appearances)
...
 				ADD_TRAIT(src, TRAIT_IMMOBILIZED, LACKING_LOCOMOTION_APPENDAGES_TRAIT)
 
+	update_limbless_movespeed_mod()
+
+/mob/living/proc/update_limbless_movespeed_mod()
 	if(usable_legs < default_num_legs)
 		var/limbless_slowdown = (default_num_legs - usable_legs) * 3
 		if(!usable_legs && usable_hands < default_num_hands)
 			limbless_slowdown += (default_num_hands - usable_hands) * 3
+		for(var/obj/item/cane/thing in held_items)
```
CRITAWAKETS, line 2299: Not a fan of this being a hardcoded thing. Maybe make it a check for traits on held items?

MrMelbert: Annoyingly it can't be as generic as a trait because any items with the trait would need to run `update_limbless_movespeed_mod` on equip / drop

PR #460, opened 2024-03-18
Reverts station trait costs, adds general budget

MrMelbert:
Strait trait costs: Good in theory, but fails in practice. 

While in theory this allows for more rare traits to shine by leaving behind some "budget" to purchase more impactful traits along side smaller ones, in practice this means the less impactful traits (already regulated by higher weights) become soft-guaranteed to roll: as they effectively combine their weights into one. 

When you have a budget of `1`, and you roll a trait that costs `0.3`, you are left with `0.7`. Now you can't buy any traits that cost `1`, but we still want to spend the rest of the budget, so we are now guaranteed to buy a bunch of things that cost `0.5` or `0.3`. This sucks in practice. 

So I reverted it. All traits are `1` cost. 

But I didn't revert it to how the system used to be. In the past, it would roll 3 numbers for `num positive`, `num neutral` and `num negative`. 
Instead, we just roll 1 big pool and then randomly pick from either positive, neutral, or negative, as we attempt to fill out or budget. 

Why am I doing this? I think it will result in more "dynamic" rounds than when it used to be. While in the past most rounds felt rather standard: You get something positive, something negative, and something neutral. Instead now, in some rounds you will feel blessed, and be given many positives, and in some rounds you will feel cursed, and be given many negatives. Or you will get the standard balanced result. Or who knows!

`Also remember this is config so it can be easily easily tweaked`

### DIFF
--- a/config/game_options.txt
+++ b/config/game_options.txt
@@ -508,19 +508,15 @@ DISALLOW_TITLE_MUSIC
 #GIVE_TUTORIALS_WITHOUT_DB
 
 ## Configuration for station traits of each type.
-## The first value (key) is the budget, or the space available to use for station traits of that type. Some take more space than others.
+## The first value (key) is the budget, or how many traits to spawn.
 ## The second value (assoc) is the weight associated with said budget compared to the rest for that type.
-POSITIVE_STATION_TRAITS 0 8
-POSITIVE_STATION_TRAITS 1 4
-POSITIVE_STATION_TRAITS 2 2
-POSITIVE_STATION_TRAITS 3 1
-
-NEUTRAL_STATION_TRAITS 0 10
-NEUTRAL_STATION_TRAITS 1 10
-NEUTRAL_STATION_TRAITS 2 3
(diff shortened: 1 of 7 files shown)


config/game_options.txt
```diff
@@ -508,19 +508,15 @@ DISALLOW_TITLE_MUSIC
...
-NEUTRAL_STATION_TRAITS 0 10
-NEUTRAL_STATION_TRAITS 1 10
-NEUTRAL_STATION_TRAITS 2 3
-NEUTRAL_STATION_TRAITS 2.5 1
-
-NEGATIVE_STATION_TRAITS 0 8
-NEGATIVE_STATION_TRAITS 1 4
-NEGATIVE_STATION_TRAITS 2 2
-NEGATIVE_STATION_TRAITS 3 1
+## Adds up to 100 for easy math: The weight = the % chance of being picked.
```
CRITAWAKETS, line 513: This should probably be rewritten to not make it seem like the combined total has to be 100.

PR #462, opened 2024-03-20
crucial tgui fix taken from tg

notfrying1pans:
taken from https://github.com/tgstation/tgstation/pull/81968, will make searching not terrible


notfrying1pans:
of note i briefly considered fixing the character prefs tooltip bug but that required a big tooltip rework pr to do and i didnt have the energy to

PR #493, opened 2024-05-19
Adds Unique Clothing For Shiro-2

Wollywoger:
## story content
this adds clothing based on pink's character, Shiro-2

Shiroclothingpicture

adds a cape and a dress.
cape's in the neck category, dress is in formal

created as a birthday gift.


Wollywoger:
i did the type path for the loadout datum differently than normal, as part of melbert's desire to have stuff in story content be able to be disabled easily.

PR #499, opened 2024-05-27
null checks surgery tables, they can be deconned now

CRITAWAKETS:
yeah thats it
just a bug fix move along citizen


maplestation_modules/code/game/objects/structures/surgery_table.dm
```diff
@@ -61,7 +61,8 @@
 
 
 /obj/structure/table/optable/deconstruct(disassembled, wrench_disassembly)
-	attached_tank.forceMove(drop_location())
+	if(!isnull(attached_tank))
+		attached_tank.forceMove(drop_location())
```
MrMelbert, line 65:
```suggestion
	attached_tank?.forceMove(drop_location())
```

PR #500, opened 2024-06-10
New Synth Customisation Options

Constellado:
Adds options for synth heads!

1. There is a new organ that goes on the head of the synth, and acts as a kind of "cover" that goes over their head, augmenting how they look!
3. I changed what the default synth head looks like, it is now like a skeleton. This head has eyes! Which you can recolour.
5. Synths also now have a pickable gender!

Without a head cover:
image
image
image

Sprites are from me apart from the IPC sprite, which is from "icons\mob\clothing\head\costume.dmi".
Helm default:
image
image
image

Helm white:
image
image
image

IPC (was from somewhere else in the code base)
(notice! It hides hats due to the huge size!)
image
image
image

Also this option!!
image
image


maplestation_modules/code/modules/mob/living/carbon/human/species_types/synth/synth.dm
```diff
@@ -447,4 +453,91 @@
 	biological_state = BIO_ROBOTIC|BIO_BLOODED
 	change_exempt_flags = BP_BLOCK_CHANGE_SPECIES
 
+/obj/item/organ/internal/eyes/robotic/synth
+	name = "synth eyes"
+
+/obj/item/organ/internal/eyes/robotic/synth/generate_body_overlay(mob/living/carbon/human/parent)
+	return ..()
```
MrMelbert, line 460: Doing nothing?

maplestation_modules/code/modules/mob/living/carbon/human/species_types/synth/synth.dm
```diff
@@ -447,4 +453,91 @@
...
+	dna_block = DNA_SYNTH_HEAD_COVER_BLOCK
+	organ_flags = ORGAN_ROBOTIC
+
+	bodypart_overlay = /datum/bodypart_overlay/mutant/synth_head_cover
+
+
+/obj/item/organ/external/synth_head_cover/on_mob_insert(mob/living/carbon/organ_owner, special, movement_flags)
+	. = ..()
+	var/mob/living/carbon/human/robot_target = organ_owner
+	var/obj/item/bodypart/head/noggin =  robot_target.get_bodypart(BODY_ZONE_HEAD)
```
MrMelbert, line 484:
```suggestion
	var/obj/item/bodypart/head/noggin = robot_target.get_bodypart(BODY_ZONE_HEAD)
```

maplestation_modules/code/modules/mob/living/carbon/human/species_types/synth/synth.dm
```diff
@@ -447,4 +453,91 @@
...
+
+	noggin.head_flags &= ~HEAD_EYESPRITES
+
+
+/obj/item/organ/external/synth_head_cover/on_mob_remove(mob/living/carbon/organ_owner, special, movement_flags)
+	. = ..()
+	var/mob/living/carbon/human/robot_target = organ_owner
+	var/obj/item/bodypart/head/noggin =  robot_target.get_bodypart(BODY_ZONE_HEAD)
```
MrMelbert, line 492:
```suggestion
	var/obj/item/bodypart/head/noggin = robot_target.get_bodypart(BODY_ZONE_HEAD)
```

maplestation_modules/code/modules/mob/living/carbon/human/species_types/synth/synth.dm
```diff
@@ -447,4 +453,91 @@
...
+
+//-- accessories --
+//the path to the icon for the head covers
+/datum/sprite_accessory/synth_head_cover
+	icon = 'maplestation_modules/icons/mob/synth_heads.dmi'
+
+//head covers
+/datum/sprite_accessory/synth_head_cover/none // for those that don't want a cover.
+	name = "None"
+	icon_state = null
```
MrMelbert, line 518: I don't think you need an bespoke type for this, I think instead all you need to do is when you make the global list
(`init_sprite_accessory_subtypes`),  you add `add_blank = TRUE` to the call, and that adds an empty sprite accessory for you

PR #506, opened 2024-06-20
Improves blood sprites, allows simplebots to track blood

MrMelbert:
- Improved the blood sprites

So the original sprites were ported from the original pr, but I had to update them, so they were kinda ugly. 
So I took inspiration from Shiptest's versions of the same sprites, which shaded them better. 
But midway through updating them manually I just said screw it and ported their versions, which were based off the original versions anyways, so all credit to the original authors and those who modified it I suppose

Ultimately they should look marginally better, some of the gibs still look a bit ass but such is life. 

I also ported their footstep smear sprites while I was in the neighborhood but I'm fairly certain those themselves were ported. If I'm being mistaken that's my bad but I thought I recognize them from the paradise-sphere. All credit to the original authors once again. 

image

- Reduced darkness of dried blood

I thought the dried blood looked *too* dim, so I brought it up slightly

image

- Footsteps should be less jank

I changed footsteps to only update alpha on creation so they should just, vanish, less often. 

- Blood drip fix

They accidentally were stacking their colors on one another which made them look generally black, that's fixed

- Allow bots, silicons, drones to track blood

Just cause I think it's cute, it reminds me of that one spessmanart piece with the malfunctioning bots

image

If one of you yells at me because mediborgs are technically levitating and don't track blood, L bozo


CRITAWAKETS:
> If one of you yells at me because mediborgs are technically levitating and don't track blood, L bozo

They're levitating through _some means_, Newton's third law says they're 100% just blasting blood everywhere if they walk above a puddle.

PR #519, opened 2024-06-29
Table Smasher (Table now Cracks with Damage, and Table Smashing now Deals Damage.)

Andrymeda:
what it says on the tin.

### DIFF
--- a/maplestation_modules/code/game/objects/structures/tables_racks.dm
+++ b/maplestation_modules/code/game/objects/structures/tables_racks.dm
@@ -5,12 +5,20 @@
 /obj/structure/table/Initialize()
 	. = ..()
 	if(!crack_states)
-	crack_states = list()
-	for(var/i in 1 to crack_states_count)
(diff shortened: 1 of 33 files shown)


maplestation_modules/code/game/objects/structures/tables_racks.dm
```diff
@@ -0,0 +1,24 @@
...
+/obj/structure/table/tablepush(mob/living/user, mob/living/pushed_mob)
+	. = ..()
+	take_damage(20)
+
+/obj/structure/table/deconstruct(disassembled, wrench_disassembly)
+	if(!disassembled)
+		playsound(src, 'sound/effects/bang.ogg', 90, TRUE)
+		visible_message(span_danger("[src] breaks down!"),
+			span_danger("You break down!"),
+			span_danger("You hear something breaking."))
```
MrMelbert, line 23: Sorry but `You break down!`, do you think someone's going to be playing the table? 😄

PR #534, opened 2024-07-14
Replaces ethereal sounds with human sounds.

CRITAWAKETS:
Oh my god, these sound SO BAD.

do not merge until after the round


MrMelbert:
I would prefer if you mixed new sounds for Ethereals which is like staticy-human sounds than make them human outright

MrMelbert:
It could probably be normalized, I bet it peaks badly

CRITAWAKETS:
Am too lazy to actually finish this PR and it's a bad approach, server has this in the tailinclude anyway

PR #536, opened 2024-07-17
Changes lobby ready up printout

MrMelbert:
The status tab in the lobby now prints all the players who are readied up.

**By default** it only shows ckey. Character and Top Job are anonymous.

image

**There is however** a preference that allows you to toggle visibility of character and top job (or your ckey if you really want). 

image

image

The intent behind this is so people can better plan around their roundstart job positions, most notably for jobs like captain or contested jobs like chaplain. 
People often discuss what they're playing before the round so might as well expedite it for those who care.


CRITAWAKETS:
I think command jobs and AI should be always visible, since they're usually contested.

PR #543, opened 2024-07-30
Polite Split Personality (non-random Split Personality Variant)

Andrymeda:
This version of Split Personality, instead of randomly changing, adds a new action shared between the two players which can be turned on or off.

If both of them are on, the players switch control on mutual demand instead.

Don't ask me why.

### DIFF
--- a/maplestation_modules/code/datums/brain_damage/split_personality.dm
+++ b/maplestation_modules/code/datums/brain_damage/split_personality.dm
@@ -34,8 +34,6 @@
 	var/datum/action/request_switch/twin_action
 	var/mob/living/split_personality/non_controller
 
-	var/switch_msg = span_boldnotice("You switch with your other personality.")
-
 /datum/action/request_switch/owner
 
 /datum/action/request_switch/stranger
--- a/maplestation_modules/code/datums/brain_damage/split_personality.dm
+++ b/maplestation_modules/code/datums/brain_damage/split_personality.dm
@@ -30,10 +30,6 @@
 
 	var/requested = FALSE
 
-	var/datum/brain_trauma/severe/split_personality/trauma
-	var/datum/action/request_switch/twin_action
-	var/mob/living/split_personality/non_controller
-
 /datum/action/request_switch/owner
 
 /datum/action/request_switch/stranger
@@ -50,11 +46,9 @@
 	if(!.)
 		return FALSE
 
-	trauma = target
-
 	requested = !requested
 
-/datum/action/request_switch/proc/handle_switch()
+/datum/action/request_switch/proc/handle_switch(datum/brain_trauma/severe/split_personality/trauma, datum/action/request_switch/twin_action)
 	trauma.switch_personalities()
 	requested = FALSE
 	twin_action.requested = FALSE
@@ -62,7 +56,11 @@
 
 /datum/action/request_switch/owner/Trigger(trigger_flags)
 	. = ..()
-	var/non_controller
+	if(!.)
+		return
+	var/datum/brain_trauma/severe/split_personality/trauma = target
+	var/datum/action/request_switch/twin_action
+	var/mob/living/split_personality/non_controller
 
 	if(trauma.current_controller == /*OWNER*/0)
 		twin_action = locate(/datum/action/request_switch) in trauma.stranger_backseat.actions
@@ -72,12 +70,18 @@
 		non_controller = trauma.owner_backseat
 
 	if(twin_action.requested)
-		handle_switch()
+		handle_switch(trauma, twin_action)
 	else
 		request(trauma.owner, non_controller)
 
 /datum/action/request_switch/stranger/Trigger(trigger_flags)
 	. = ..()
+	if(!.)
+		return
+
+	var/datum/brain_trauma/severe/split_personality/trauma = target
+	var/datum/action/request_switch/twin_action
+	var/mob/living/split_personality/non_controller
 
 	twin_action = locate(/datum/action/request_switch)  in trauma.owner.actions
 	if(trauma.current_controller == /*OWNER*/0)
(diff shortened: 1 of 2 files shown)


maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,124 @@
...
+
+/datum/action/request_switch/proc/handle_switch()
+	trauma.switch_personalities()
+	requested = FALSE
+	twin_action.requested = FALSE
+	return
+
+/datum/action/request_switch/owner/Trigger(trigger_flags)
+	. = ..()
+	var/non_controller
```
MrMelbert, line 65: What is this `var/non_controller` doing here?

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,124 @@
...
+		twin_action = locate(/datum/action/request_switch) in trauma.owner_backseat.actions
+		non_controller = trauma.owner_backseat
+
+	if(twin_action.requested)
+		handle_switch()
+	else
+		request(trauma.owner, non_controller)
+
+/datum/action/request_switch/stranger/Trigger(trigger_flags)
+	. = ..()
```
MrMelbert, line 80:
```suggestion
	. = ..()
	if(!.)
		return
```

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,124 @@
...
+	background_icon_state = "bg_spell"
+	button_icon = 'icons/mob/actions/actions_spells.dmi'
+	button_icon_state = "swap"
+	overlay_icon_state = "bg_spell_border"
+
+	var/requested = FALSE
+
+	var/datum/brain_trauma/severe/split_personality/trauma
+	var/datum/action/request_switch/twin_action
+	var/mob/living/split_personality/non_controller
```
MrMelbert, line 35: For all of these, you need to set `[blar] = null` in `/datum/action/request_switch/Destroy`

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,124 @@
...
+	if(!istype(target, /datum/brain_trauma/severe/split_personality))
+		stack_trace("[type] was created on a target that isn't a /datum/brain_trauma/severe/split_personality, this doesn't work.")
+		qdel(src)
+
+/datum/action/request_switch/Trigger(trigger_flags)
+	. = ..()
+	if(!.)
+		return FALSE
+
+	trauma = target
```
MrMelbert, line 53: Why is this in `Trigger`?

Actually you don't need this at all, if you need to cast trauma you can just do it in whatever proc it's necessary:
`var/datum/brain_trauma/severe/split_personality/trauma = target`

maplestation_modules/code/datums/brain_damage/split_personality.dm
MrMelbert: Actually, I don't think you need any of these on the datum

You should be able to put them in the proc it's used?

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,128 @@
...
+	if(!.)
+		return
+	var/datum/brain_trauma/severe/split_personality/trauma = target
+	var/datum/action/request_switch/twin_action
+	var/mob/living/split_personality/non_controller
+
+	if(trauma.current_controller == /*OWNER*/0)
+		twin_action = locate(/datum/action/request_switch) in trauma.stranger_backseat.actions
+		non_controller = trauma.stranger_backseat
+	else if(trauma.current_controller == /*STRANGER*/1)
```
SmArtKar, line 68: Should probably be #define PERSONALITY_SWITCH_OWNER 0 and #define PERSONALITY_SWITCH_STRANGER 1 instead of comments

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,128 @@
...
+
+/datum/action/request_switch/stranger/proc/request(mob/controller, mob/non_controller)
+	var/str_request_msg = span_boldnotice("You concentrate on attempting to take direct control over your body.")
+	var/str_requested_msg = span_boldnotice("You feel your other side attempting to take direct control over your body.")
+
+	var/str_cancel_msg = span_boldnotice("You give up on attempting to take direct control over your body.")
+	var/str_cancelled_msg = span_boldnotice("You feel your other side giving up on taking direct control over your body.")
+
+	controller.balloon_alert(controller, "your brain tickles")
+	non_controller.balloon_alert(controller, "your brain tickles")
```
SmArtKar, line 105: This should probably be 
```
controller.balloon_alert(controller, "your brain tickles")
controller.balloon_alert(non_controller, "your brain tickles")
```

maplestation_modules/code/datums/brain_damage/split_personality.dm
```diff
@@ -0,0 +1,128 @@
...
+
+/datum/action/request_switch/owner/proc/request(mob/controller, mob/non_controller)
+	var/own_request_msg = span_boldnotice("You concentrate on giving your other side control to your body.")
+	var/own_requested_msg = span_boldnotice("You feel your other side attempting to let you take over their body.")
+
+	var/own_cancel_msg = span_boldnotice("You give up on attempting to give your other side control to your body.")
+	var/own_cancelled_msg = span_boldnotice("You feel your other side giving up on giving you control.")
+
+	controller.balloon_alert(controller, "your brain tickles")
+	non_controller.balloon_alert(controller, "your brain tickles")
```
SmArtKar, line 122: same here

SmArtKar (changes requested):

PR #558, opened 2024-08-11
Crutches + Allows mobs with 0 usable legs to walk with 2 canes

MrMelbert:
- Cherry-picks the crutch + craft-able crutch from /tg/ upstream
- Canes help with limps, either pain or wound (as they do upstream) 
- Mobs with no usable legs can now walk if they have a crutch/cane in both hands
- Mobs with no usable legs should not be able to float around in no gravity more consistently

Something to note, if you've propped yourself up with two canes - the moment you drop one, you drop to the floor. 

I can see this getting a little annoying, so maybe we need to follow this up with letting people prop themselves against walls or tables (so you can do something with your free hand before continuing to move) (maybe utilize the existing lean system?)

Also we can probably follow this up with a preference for paraplegics to let them choose between wheelchair or crutches


MrMelbert:
Updated with 
- a preference for paraplegics to spawn with crutches rather than wheelchairs 
- leaning against the wall counts as 1 support out of 2 supports needed to stand with non-functioning legs (so 1 crutch + wall works)
- you can brace yourself on a crutch to have it count as 2 supports needed to stand, allowing you to stand with 1 crutch and use a free hand (but you lose effect this if you take a step)

PR #559, opened 2024-08-11
Get some shut-eye

MrMelbert:
If you fall asleep/unconscious (or die), your eyes will close. 

image

This does not apply to robotic eyes (they will always be wide open) or eyes that don't have color (like moth eyes - it would be better suited to give those unique shut sprites)


CRITAWAKETS:
Why is this not upstream aswell? It seems like it'd be a good & fitting addition there aswell.

MrMelbert:
I'll probably do it later

PR #565, opened 2024-08-18
Fix mauling

MrMelbert:
Melee cd is not handled in melee, but before melee. Which we skip here.


Bluedino1025:
context https://www.youtube.com/watch?v=8rWAA-WwQcs

PR #573, opened 2024-08-26
Adds defenestration and bullet penetrating glass

MrMelbert:
1.

If a bullet would fully shatter a window, there's an 80% chance the shot will continue.

2.

Being tossed against a (non-reinforced) window now does considerably more damage (to the window)
By default, normal windows require a human to be chucked at it twice to break.

If the toss would shatter the window, you would instead fly _through_ the window, assuming it has no grille. 
This does additional damage (to you) and may cause the shards to embed.

(Demo, has since been tweaked)

https://github.com/user-attachments/assets/0d6156f0-ecba-4799-9eea-805d8272a197

### DIFF
--- a/code/datums/components/caltrop.dm
+++ b/code/datums/components/caltrop.dm
@@ -78,7 +78,7 @@
 	if((flags & CALTROP_IGNORE_WALKERS) && digitigrade_fan.move_intent == MOVE_INTENT_WALK)
 		return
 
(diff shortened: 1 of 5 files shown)


MrMelbert:
Defenestration will be a lot cooler when Wallening actually rolls around and we get window bases rather than grilles. Until then it's a little lame (seeing as most windows have a grille)

PR #583, opened 2024-09-12
Redoes Bodytemp [MDB Ignore]

MrMelbert:
Basically rewrites body temperature entirely, so people accurately homeostasis their body temperature to normal amounts (IE: not 21c), you can feel cold without immediately taking damage, lizards are not fricked up, etc.


MrMelbert:
We can test this now, I'm sure I missed some things but in average station living it seems fine

code/modules/mob/status_procs.dm
```diff
@@ -8,11 +8,126 @@
...
+ * * temperature - what temperature is being exposed to this mob?
+ * some articles of clothing are only effective within a certain temperature range
+ *
+ * returns the percentage of protection as a value from 0 - 1
+**/
+/mob/living/proc/get_insulation(temperature = T20C)
+	// There is an occasional bug where the temperature is miscalculated in areas with small amounts of gas.
+	// This is necessary to ensure that does not affect this calculation.
+	// Space's temperature is 2.7K and most suits that are intended to protect against any cold, protect down to 2.0K.
+	temperature = max(temperature, T0C)
```
dwasint, line 33: T0C is 270k this means everything covers space, should be TCMB

dwasint:
adjust_body_temperature seems fundamentally flawed, as you will eventually always equalize with the rooms ambient temperature assuming you aren't at 1.0 insulation. For environmental changes you'd need to probably create an equalization point based on the areas temperature as say icebox should equalize at around 5 celcius with 0.75 insulation. However you will eventually reach iceboxes outside temperature. This could be turned into an intended feature but it essentially blacklists lizards from mining pre-modsuit or anyone really as they will start freezing in like 5 minutes

dwasint:
> adjust_body_temperature seems fundamentally flawed, as you will eventually always equalize with the rooms ambient temperature assuming you aren't at 1.0 insulation. For environmental changes you'd need to probably create an equalization point based on the areas temperature as say icebox should equalize at around 5 celcius with 0.75 insulation. However you will eventually reach iceboxes outside temperature. This could be turned into an intended feature but it essentially blacklists lizards from mining pre-modsuit or anyone really as they will start freezing in like 5 minutes

actually the issue would be with handle_environment it should probably check for equalized temperature

dwasint:
/mob/living/proc/get_insulated_equilibrium_temperature(environment_temp, insulation)
    return environment_temp + (standard_body_temperature - environment_temp) * insulation
    
    should work for pulling the insulated equalibrium temp for handle_environment

MrMelbert:
> as you will eventually always equalize with the rooms ambient temperature assuming you aren't at 1.0 insulation.

i'm not sure what you mean by this exactly, humans are supposed to equalize with the room's temperature, their homeostasis is supposed to offset it (which it's strong enough to do around room temperature, but not strong enough in, say, icebox's exterior) 

are you saying it equalizes too _fast_ in places like icebox?

> This could be turned into an intended feature but it essentially blacklists lizards from mining pre-modsuit or anyone really as they will start freezing in like 5 minutes

the mining suit is 100% insulated so it's fine (though i had to go out of my way to add boot sprites for lizards so they could mine safely)

MrMelbert:
i i think i see what you mean, any amount of insulation below 1 just slows you from reaching area temp rather than keep from above it. i'll play around with it but i wouldn't want it to be 100% perfect, i imagine it'd mess with some thresholds since a silly amount of things have some amount insulation on it (gloves namely)

MrMelbert:
> /mob/living/proc/get_insulated_equilibrium_temperature(environment_temp, insulation) return environment_temp + (standard_body_temperature - environment_temp) * insulation

this equation definitely feels too aggressive about it, as a single winter coat with its hood makes you immune to icebox indefinitely, which i wanted to avoid. i'll try tweaking it

dwasint:
> i i think i see what you mean, any amount of insulation below 1 just slows you from reaching area temp rather than keep from above it. i'll play around with it but i wouldn't want it to be 100% perfect, i imagine it'd mess with some thresholds since a silly amount of things have some amount insulation on it (gloves namely)

its mainly because even with a coat and boots on a lizard they will start freezing in around 30 seconds, they need 100% insulation to exist outside for more then a minute coats should ideally slow it drastically

dwasint:
also yea it would protect against heat but thats because we don't have heat insulation and freezing insulation

MrMelbert:
> its mainly because even with a coat and boots on a lizard they will start freezing in around 30 seconds, they need 100% insulation to exist outside for more then a minute coats should ideally slow it drastically

A full mining outfit is 100%, gloves-boot-suit-hood(up)

dwasint:
> > its mainly because even with a coat and boots on a lizard they will start freezing in around 30 seconds, they need 100% insulation to exist outside for more then a minute coats should ideally slow it drastically
> 
> A full mining outfit is 100%, gloves-boot-suit-hood(up)

don't think the ice bitmining domain has gloves tho so they just freeze

MrMelbert:
probably need to fix that

MrMelbert:
Yolo

PR #590, opened 2024-09-25
Adds Combat Gloves Single-Pack.

CRITAWAKETS:
image

This just allows you to buy combat gloves as a goodie for a decently large price. Probably too large of a price, but that's what you get for wanting DRIP (that looks exactly the same as black gloves)

These USED to be sold in cargo, but the options for buying them withered away until they were no longer available.


Wallemations:
Don't combat gloves give a pretty decent fire/bio armor boost?

CRITAWAKETS:
Fire armor's only used for the actual clothing's durability to being burnt. It's a boost but not a major one.
Bio armor is actually used for things, preventing disease, making you more resistant to bio-scramblers and get exposed less to reagents. That said, haven't seen many people exactly run for bio resistance that isn't 100% effective.

PR #594, opened 2024-09-30
A special character: Localnode#4248

Constellado:
## What is this?

Localnode is a character that Blacklist897 controls. It is a handheld AI-like character which can speak as itself when using it like a radio!
image

### Right hand sprites:
imageimageimageimage

### Left hand sprites:
imageimageimageimage


Blacklist897:
thank you very much bolkan

PR #603, opened 2024-10-21
Adds more languages to the AI.

Constellado:
# What is this?
This code adds more languages to the AI. 

## Why?
I made this so CaLE can understand and speak more languages, it was missing quite a few.

The languages I chose were ones that CaLE would know IC at this current point of time. If adding yangyu is a bit much, tell me and I will remove it.

Languages added:
1. Draconic
2. Yangyu


lizelive:
bogos shouldn't know more languages. split?

PR #619, opened 2024-11-14
Adds transparent ends to tile color turf decals.

CRITAWAKETS:
This PR adds in transparent ends to the turf decals that are used to color tiles, aswell as trimlines. This therefore extends them to the edges of the tile.

Very minor change but one that my eyeballs want so, so much.

Example of before and after:
image

The Meta Armory circles are less lop-sided now:
image

Example of Meta Bridge:
image

Is that a red smooth tile that doesn't have weird green ends sticking out of it??
image


CRITAWAKETS:
Sadly after testing, this doesn't really work too well. Decals often look like they're "sunken in" compared to other decals because of the shading being obscured and it's likely attempting to work around this problem would cause many more issues.

also thinking byond's broken-ass alpha rendering might have a thing to play here, so uh. oops. byond issue!

PR #645, opened 2025-01-12
Umbrellas now shade shadowpeople from light.

CRITAWAKETS:
This PR makes umbrellas, once active, shade shadowpeople from light. What this means is that they don't take light damage, but they don't get the shadow regeneration either unless they're in the dark.

This PR also adds in umbrellas as a crafting recipe and in the clothesmate, to add in a non-loadout source.

Currently this PR does not add in automatic umbrella spawning for shadowpeople joining in roundstart, because that's a complicated mess of code that's going to take more time to make and test than the rest of the PR.


maplestation_modules/code/game/objects/items/umbrellas.dm
```diff
@@ -88,6 +88,10 @@
 	open = active
 	if(user)
 		balloon_alert(user, active ? "opened" : "closed")
+	if(active)
+		ADD_TRAIT(user, TRAIT_SHADED, UMBRELLA_TRAIT)
+	else
+		REMOVE_TRAIT(user, TRAIT_SHADED, UMBRELLA_TRAIT)
```
MrMelbert, line 94: You need to remove the trait in `dropped` and add the trait in `equipped` (if active), else you could just extend the umbrella and drop it and have shade forever

code/__DEFINES/traits/declarations.dm
```diff
@@ -222,6 +222,9 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 ///Added to mob or mind, changes the icons of the fish shown in the minigame UI depending on the possible reward.
 #define TRAIT_REVEAL_FISH "reveal_fish"
 
+// Trait that says you're shaded by something, used on shadowpeople to prevent DoT from light
+#define TRAIT_SHADED "shaded"
```
MrMelbert, line 226:
```suggestion
/// Trait that says you're shaded by something (ie in the dark) 
#define TRAIT_SHADED "shaded"
```
Should generally avoid saying the precise reason a trait exists, as other things can piggyback off the trait if they want

maplestation_modules/code/game/objects/items/umbrellas.dm
CRITAWAKETS: not my ass forgetting this logic

PR #656, opened 2025-01-30
Moves init time display out of chat (and make it hideable since most people don't care)

MrMelbert:
Init time is displayed on text that scrolls on the left side of the lobby screen

The text will fade out 30 seconds to roundstart

There is also a preference to hide it entirely, so you can enjoy the lobby art.

image


MrMelbert:
I slash we may want to make this diagetic, like the ready screen, but I can't sprote it so

PR #660, opened 2025-02-08
Heath rework patch one

MrMelbert:
Changes:

- The rate at which coldness slows you down has been reduced by 5x
   - This also means the maximum amount of slowdown from coldness is 5x less

- Tweak to lung damage from vacuums
   - Halved the damage per breath

- Tweaks to shaft miners (namely their skill chip)
   - While on lavaland, shaft miners can sprint 1.5x longer, and their sprint regenerates 1.5x faster
   - Additionally, they have access to an "emergency sprint", which injects them with a few chemicals that boosts their runspeed and nullifies slowdown from damage. 
   - These are intended help people escape from megafauna they accidentally aggro, since - as it turns out - accidentally sprinting head-first into one was pretty deadly.

- Adds a VV dropdown to humans called "make me tanky" which admins can use to artificially make a human more resilient 

- Ports + tweaks to medical items
   - Auto-cycling between damaged limbs on LMB
   - Most medical items now work on pierce immune skin

- Tweaks to silverscales
   - Tongues are no longer worth 65000 points, now instead worth 200 
   - Tongue no longer grants `TRAIT_PIERCEIMMUNE`, `TRAIT_RESISTHIGHPRESSURE`, `TRAIT_RESISTLOWPRESSURE`, `TRAIT_VIRUSIMMUNE`. I intend to re-integrate these into the tongue later on, but I have to refactor bodyparts a bit first

- Tweaks to CPR
   - When you help-intent click on someone who is on the floor, it now opens a radial. You can pick from checking their pulse, doing CPR, or just shaking them.
   - This behavior can be toggled off in the game settings
   - There is two emotes, *cpr and *check which allow you to do those actions if you toggle off the help radial.  

- Vitals floor scanner can be rotated 

- Adds Nitroglycerin pills to cargo prescriptions

Planned changes:

- Tweaks to shock and paincrit to make it less binary
- Tweaks to nobreath 
- Tweaks to the heartbeat SFX to make it less constant
- Re-adding feeling "cold" when you blood is low (and general reworks to blood volume)
- Adjusting damage / effects from high temp

Distant changes:

- Reworks to being husked (making it per-bodypart)
- Reworks to hypo / hyperthermia
- Adding proper splints
- Frostbite


code/modules/mob/living/carbon/human/human.dm
```diff
@@ -857,7 +857,62 @@
...
 
+		var/major = tgui_input_list(usr, "REALLY tanky or just a little tanky?", "Tanky", list("Little", "Lot"))
+		if(QDELETED(src) || !major)
+			return
+		major = (major == "Lot")
+
+		for(var/i in picked)
+			switch(i)
+				if("More effective HP")
+					add_consciousness_modifier("badmin", major ? 30 : 10)
```
CRITAWAKETS, line 890: Applies to the whole of it, but the exact changes in stats should be catalogued somewhere out of code to make it easier for newer admemes.

code/modules/mob/living/carbon/human/human.dm
```diff
@@ -857,7 +857,62 @@
 
 		qdel(src)
 
+	if(href_list[VV_HK_MAKE_ME_TANKY])
```
CRITAWAKETS, line 860: Needs to be logged, if unlogged this would be a MAJOR route for possible admin abuse shenanigans.

code/modules/surgery/organs/internal/lungs/_lungs.dm
```diff
@@ -655,7 +655,7 @@
 				to_chat(breather, span_boldwarning("You feel air force itself into your lungs!"))
 
 			breather.cause_pain(BODY_ZONE_CHEST, 10, BRUTE)
-			apply_organ_damage(15)
+			apply_organ_damage(maxHealth * 0.075)
```
CRITAWAKETS, line 658: Is this intended to always have the same amount of damage based on the lung? This would really only affect smokers right now but i'm not sure about max health-based scaling on this

maplestation_modules/code/modules/library/skill_learning/job_skillchips/mining.dm
```diff
@@ -3,20 +3,43 @@
...
 	skill_icon = "fist-raised"
 	activate_message = span_notice("You feel like you can safely take on the unknown.")
 	deactivate_message = span_notice("You feel more vulnerable to the unknown.")
 
+	var/datum/action/cooldown/miner_skillchip/action
+	var/refill_cd = 10 MINUTES
+
+/obj/item/skillchip/job/off_z_pain_resistance/Initialize(mapload, is_removable)
+	. = ..()
+	create_reagents(25)
```
CRITAWAKETS, line 19:
```suggestion
	create_reagents(max_vol = 25)
```

code/modules/surgery/organs/internal/lungs/_lungs.dm
MrMelbert: Figured it'd be fairer

maplestation_modules/code/modules/library/skill_learning/job_skillchips/mining.dm
MrMelbert: You'll never catch me alive

PR #662, opened 2025-02-12
The alt job title thing + Honorifics

MrMelbert:
UX was inspired by that seen on insert-furry-server-of-choice, but implementation is my own. 

Specifically it lets different titles have different roundstart outfits, inspired by a conversation in coderbus I saw. So "Surgeons" start in full scrubs while "Nurses" start in a normal jumpsuit. 

image

Maybe we move Xenobiologist and Ordnance to this system, maybe not.
Conceptually I still like having specialist job types, like Virologist, Xenobiologist, Atmospheric Technician. I think so long as a job has a defined workspace it should be its own job rather than an alt title. ("Surgeon", for example, works in the same space as Medical Doctors do, ultimately.) 

Also I don't want to spam this with a bunch of super niche titles. Like titles that are just "a more fancy or less fancy way of saying the same thing".

### DIFF
--- a/code/__DEFINES/id_cards.dm
+++ b/code/__DEFINES/id_cards.dm
@@ -36,3 +36,28 @@
 )
 /// Wildcard slot define for admin/debug/weird, special abstract cards. Can hold infinite of any access.
 #define WILDCARD_LIMIT_ADMIN list(WILDCARD_NAME_ALL = list(limit = -1, usage = list()))
+
+/**
+ * x1, y1, x2, y2 - Represents the bounding box for the ID card's non-transparent portion of its various icon_states.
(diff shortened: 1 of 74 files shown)


Wallemations:
> Maybe we move Xenobiologist and Ordnance to this system, maybe not.

One thing that confuses me whenever I play a scientist is whether or not I'm just one of the other roles with less access, so it might be a good idea tbh

MrMelbert:
Scientist is just "those roles with less access"

Wallemations:
> Scientist is just "those roles with less access"
> 
> 
> 
> They're a parallel to Virologist

Damn they don't get _any_ extra stuff at all?

PR #659, opened 2025-02-05
Headset circuit shells output their wearer

Wallemations:
What the tin says

Honestly here mostly for a more personalized translator design I've stumbled across that I liked a lot lmao

PR #676, opened 2025-02-24
Circuit headsets now ACTUALLY output their current user & Module Components can be linked

Wallemations:
I messed up #659 and the circuit headset wasn't actually doing the thing I said it should have done. Now they do!

Also I finally got around to stealing the part of #550 that made Module Components able to be linked to component printers, since that turned out to actually not be all that much code at all


Wallemations:
I must have accidentally left drafting on from my last PR, oops

PR #693, opened 2025-03-24
Gives Stabilizer chromosomes a pity chance

Wallemations:
if you don't get one it adds +5 to the weight

I haven't tested it yet so draft


Wallemations:
tested, it works

PR #708, opened 2025-04-21
Assuage crashing

MrMelbert:
We got the bad seed


CRITAWAKETS:
i spawned in a lava lake this seed SUCKS bro

PR #710, opened 2025-04-27
adds a loser

doshMobile:
adds hair and outfit for a guy im like speedrunning something so if anything happens im gonna turn inside out


doshMobile:
oh my god im so smart hang on

PR #713, opened 2025-04-30
Detective Honorific: Sheriff

Wallemations:
Yeehaw!!!!!!

Idk if I set this up right lmao

### DIFF
--- a/code/datums/id_trim/jobs.dm
+++ b/code/datums/id_trim/jobs.dm
@@ -570,7 +570,7 @@
 		ACCESS_HOS,
 	)
 	job = /datum/job/detective
-	honorifics = list("Detective", "Investigator")
+	honorifics = list("Detective", "Investigator", "Sheriff")
 	honorific_positions = HONORIFIC_POSITION_FIRST | HONORIFIC_POSITION_LAST | HONORIFIC_POSITION_FIRST_FULL
 
 /datum/id_trim/job/detective/refresh_trim_access()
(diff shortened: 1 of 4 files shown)


code/datums/id_trim/jobs.dm
```diff
@@ -583,6 +583,9 @@
 	if(CONFIG_GET(flag/security_has_maint_access))
 		access |= list(ACCESS_MAINT_TUNNELS)
 
+/datum/id_trim/job/medical_doctor/sheriff
```
MrMelbert, line 586: `/datum/id_trim/job/medical_doctor` yeah

code/modules/jobs/job_types/detective.dm
```diff
@@ -91,3 +92,10 @@
 
 	if(visualsOnly)
 		return
+
+/datum/outfit/job/detective/sheriff
+	name = "Sheriff"
```
MrMelbert, line 97: You didn't set the `id_trim` in this outtfit

code/datums/id_trim/jobs.dm
Wallemations: Oops

code/modules/jobs/job_types/detective.dm
Wallemations: Wouldn't it inherit it from parent?

MrMelbert: Yes, but don't you want to use the custom id trim you made?

Wallemations: Oh I'm stupid
Right I'll fix that

PR #717, opened 2025-05-17
Autotomy patch / Medipen fix

MrMelbert:
Autotomy reduces the amount of pain received from losing limbs - this applies to all limb loss instead of just from manually dropping limbs

Fixes copypasta with painkiller medipens


CRITAWAKETS:
using autotomy + synchronizer + ligament hook to get truly RICH on mediborg's amputation adventure!!!

PR #724, opened 2025-06-11
Big PDA and lore item fix.

Constellado:
## What is this?
This adds the big PDA that CaLE made for Airi and renames Volkan's umbrella to radiation shielded umbrella.

The big PDA has screens!

image

It works ingame :)

## Why?
I decided that coding in the thing I sprited for last round is going to be easier for admins to spawn again.

I know it may not ever show up again in lore, but I felt like taking the effort.

This also fixes the umbrella's name. It's important that it does not have Volkan's name in case somebody chooses that umbrella again.


Constellado:
Oh its mad at me for making the greyscale config modular...

Constellado:
I just did what it told, it should be good now.

PR #729, opened 2025-06-19
Makes a beam between handtele portals when you go through them!

Constellado:
# What's this??
teleport

When you go through a portal an effect between the two portals happen! You will now be able to see where the teleport is going when somebody else goes through it and you will be able to see teleports through the station as they happen.

## Why?

I had this idea while talking to people about hand tele nerfs in the discord. As you will be able to see where other players are going, It helps a lot with game visibility. It also allows for players to have a general idea of teleporters moving through the station, which helps with making the station feel a bit more busy!

I think it's a positive change however it may not be. The effect animation may also need to be adjusted/changed. 

This PR will need feedback!


CRITAWAKETS:
that actually looks a lot better than i thought it would, nice beam sprite by the way

PR #739, opened 2025-07-06
Fixes the experiscanner

Wallemations:
I wanted to fix circuits too but that PR has dependencies, this one works fine though I tested it.


Wallemations:
The only change here was a UI file so there shouldn't be any reason that the integration tests having issues with fire should have to do with my PR

Wallemations:
prettier apparently hates it. It's outputting so many red lines saying stuff only works in typescript files.

Wallemations:
ok I give up, I have no clue what I'm doing since it's gotta do with UI stuff.

PR #757, opened 2025-08-13
[CHERRY-PICK] Repaths sprite accessory lists to be on a subsystem instead of global

Wollywoger:
This is a cherry pick of https://github.com/tgstation/tgstation/pull/82847 with attempts to update our own stuff to it as well. I am doing this as part of a later cherry pick from timegreen and ghommie which kills off mutant, internal, and external organs

This is being done first as it is not only a dependency, but a notable refactor, and i'd like to save the headache of juggling refactor errors
mirroring the original, this should have no visual changes


Wollywoger:
oh boy more conflicts

PR #764, opened 2025-08-19
Fixes Rootbread soup and stops rootdough losing its reagents

DaydreamIQ:
Couple of recipes I noticed had already been fixed on live hadn't caught up here yet, makes it so you can craft a honey sweetroll and actually eat the buggers


DaydreamIQ:
Oh, dammit. Was not aware of that, also I didn't modular it so uh...ah well

PR #792, opened 2025-09-22
Replaces some door + shutter combinations with access panels

MrMelbert:
(I think I subconsciously stole this idea from Kapu) 

Some secure areas with shutters have had their shutters removed and replaced with Access Panels

image

These panels bolt shut the corresponding door until you either
1. Swipe a keycard with the correct access
2. Enter the correct password 

Upon which the door will be bolted open until you either
1. Swipe your keycard once again
2. Clear the password

Some jobs have intrinsic knowledge of the passwords for some doors
image

Some access panels may have additional requirements to be used, such as the armory panel requiring a certain alert level

AIs can override the door as normal, and the doors are hackable as normal

These serve multiple purposes:
- Ever wanted a door to be public access but had no way to accomplish that? Now you can just swipe and forget!
- Lost your ID? You can still access critical rooms for emergencies!
- Only guy who can access something on the other side of the station but you need in urgently? Ask for the password over comms!
- Mind reader? Steal the password from someone's memories!
- RP opportunities by accidentally leaving an important door open? Sure why not!

And in the future we could do stuff like
- Check a keypad for fingerprints to find the password
- Have keypad locked dorms rooms instead of a bolt

### DIFF
--- a/_maps/map_files/KiloStation/KiloStation.dmm
+++ b/_maps/map_files/KiloStation/KiloStation.dmm
@@ -2602,14 +2602,15 @@
 "anw" = (
 /obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/engineering{
-	name = "Tech Storage"
+	name = "Tech Storage";
+	id_tag = "TECH_STORAGE_PASSWORD_PANEL"
 	},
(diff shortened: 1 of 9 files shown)


MrMelbert:
You will be pleased to hear I left shutters on rooms which actually had room for them (ie, any room with 2 tile shutters)

PR #793, opened 2025-09-26
Replaces Security Disablers with Phasers (an in-between of Disablers and Energy Guns) 

MrMelbert:
Problem statement:

- Security more often than not needs to respond to threats which cannot be disabled, be it extra-dimensional gods or carp.
- These kind of threats are pretty spontaneous, and while it may be *possible* that security can book it to sec for an energy gun between the threat's warning and arrival, this is unlikely given how the server typically plays. 
- Beating these threats to death with harmbatons is simply not viable and may more often than not result in an untimely demise. 

Potential solutions:

- Do nothing. Make them go to the armory when they need a gun.
   - Going all the way to security to pick up an energy gun is not really an option. Usually that Thing needs to die Now.

- Give them energy guns roundstart.
   - Could result in security unfairly dumping lasers into members of the crew who could otherwise be disabled. 

- Give them special energy guns which have lethal mode locked to some alert level.
   - Still makes someone run across the station, just not the officer. 

Opted for solution:

- Secuity's disablers are now phasers.
- Phasers have 20 disable shots or 10 phaser shots (compared to an energy guns 20 disabler / 12 laser)
- Phaser shots are green lasers which worse damage fall off and have a whopping -50 wound penalty / 0 bare wound bonus

This gives security a pocket lethal option that's better in close range and not very effective against humanoids (it will almost never wound). In other words, it gives them lethal option that's forgiving for crewmembers who get slapped by it, yet still effective at stopping threats like carp. 

Asset Protection officers are given a Phaser as well. 
Bridge Officers retain their Disabler.
Disablers can still be found in the armory and cargo, and take on the role of the designated self defense weapon (whereas Phasers become the designated field weapon, Laser Guns exist as the designated "Big Thing Must Die" weapon, and Energy Guns exist as an upgrade to Phasers for emergencies that warrant extra firepower or for heads of staff)

### DIFF
--- a/code/modules/jobs/job_types/security_officer.dm
+++ b/code/modules/jobs/job_types/security_officer.dm
@@ -212,7 +212,6 @@ GLOBAL_LIST_EMPTY(security_officer_distribution)
 	id_trim = /datum/id_trim/job/security_officer
 	uniform = /obj/item/clothing/under/rank/security/officer
 	suit = /obj/item/clothing/suit/armor/vest/alt/sec
-	// suit_store = /obj/item/gun/energy/disabler
 	suit_store = /obj/item/gun/energy/disabler/phaser
 	backpack_contents = list(
 		/obj/item/evidencebag = 1,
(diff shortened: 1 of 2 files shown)


CRITAWAKETS:
Phasers not wounding is actually pretty great for the aspect of what they are, it means medical can easily revive the corpses provided by them.

also im a big fan of doppler's solution to this problem (give security swords) but it's also pretty silly and this fits the themes of ss13 more

code/modules/projectiles/ammunition/energy/stun.dm
```diff
@@ -24,6 +24,9 @@
 	harmful = FALSE
 	firing_effect_type = /obj/effect/temp_visual/dir_setting/firing_effect/blue
 
+/obj/item/ammo_casing/energy/disabler/phaser
+	// select_name = "stun" // maybe...
```
CRITAWAKETS, line 28: Do it.

PR #797, opened 2025-09-28
[IDB IGNORE] [MDB IGNORE] Ending External/Internal Organ Divide: Act 1

Wollywoger:
wew lad
so this does some cherry picks from tg which include 
https://github.com/tgstation/tgstation/pull/82435
https://github.com/tgstation/tgstation/pull/83744
https://github.com/tgstation/tgstation/pull/85137
some of these may seem random but they're here as dependencies for Reasons

this will be seperate from a later patch will do a find/replace to remove the typings, especially as that will be a **conflict nightmare**

Current Issues:
- [ ] Cybernetic Cat Ears Don't Work
- [ ] Cat Ear Type Vanished
- [ ] Kobolds Untested

When this is merged i'll run the find and replace, which will punt a good solid chunk of PRs here, so be ready (though for all but a select few its just removing paths)


Wollywoger:
note: theres a bug here that will be fixed fully with a future PR:
I had to forcibly set ornithid plumage ear randomization to only have one option else we get tom foolery with screenshot tests
melbert has a PR up on tgstation at the time of writing that *should* fix this, so i'll add that in the next part

PR #813, opened 2025-10-15
Restraint rebalancing

MrMelbert:
- Progress on resisting out of restraints is now saved across resist attempts 
   - Resisting will start ticking up the time to drop the restraints
   - Moving or being moved will stop resist progress from increasing
   - Being stunned, knocked down, etc. will bulk regress your escape progress equal to the duration of the effect, or half the duration if you are not actively resisting
      - Being in soft crit or unconscious will slowly tick down progress 
      - There is no other way to regress escape progress outside of re-applying the restraints
   - Being buckle-cuffed will double the time it takes to resist out of cuffs, but you no longer need to resist out of buckle, then cuffs
   - Being actively grabbed will also double the time it takes to resist out of cuffs

- Successfully breaking out of cuffs will no longer immediately remove the cuffs, but instead allows the player to drop the cuffs whenever they feel like

- You can now see someone actively resisting restraints in examine

- Legcuffs are now shown in examine 

- Improved wording of handcuff items in examine

### Why?

Well, handcuff GBJ is maybe my least favorite things ever. So this change seeks to amend that: 

- Escaping long-timer objects like straight jackets is more manageable over time
- Moving someone 1 inch every 40 seconds to keep them cuffed is no longer an option unless you want to stunlock them or move them CONSTANTLY, which requires considerably more effort. 
- Security are now incentivized to place detainees in a cell instead of stuck to a chair
- Antags are now incentivized to get their kidnap-ee to their destination with some urgency

### DIFF
--- a/code/game/objects/buckling.dm
+++ b/code/game/objects/buckling.dm
@@ -327,13 +327,13 @@
 /// Feedback displayed to nearby players after a mob is buckled to src.
 /atom/movable/proc/buckle_feedback(mob/living/being_buckled, mob/buckler)
 	if(being_buckled == buckler)
-		buckler.visible_message(
+		being_buckled.visible_message(
 			span_notice("[buckler] buckles [buckler.p_them()]self to [src]."),
 			span_notice("You buckle yourself to [src]."),
 			span_hear("You hear metal clanking."),
 		)
 	else
-		buckler.visible_message(
+		being_buckled.visible_message(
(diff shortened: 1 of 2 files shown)


CRITAWAKETS:
I feel like the knocking down and stunning effects are just going to lead to your average security cuffing experience being severe amounts of police brutality. Not sure if that's intentional.

PR #842, opened 2025-11-21
Some AI love

MrMelbert:
1. Ais can examine
2. Ais hear sounds near their AI eye if the sound is within 5 tiles of an active and broadcasting intercom. This sound is pitched down a bit and has reverb disabled so you can differentiate it from sounds actually near your core. 
3. Ais now see a ghost of their hologram where their eye is


Constellado:
YAAAAAAAY

PR #854, opened 2025-12-11
Android and Synthetic rework

MrMelbert:
Androids are no longer just "Fully augmented humans but with a bunch of arbitrary immunities" 

Androids now have an innate set of organs which modify their core functions. Some of these concepts were vaguely inspired by some things Goofball were working on

### Bobots

- These Android organs are unremovable, at least for now. No upgrades allowed.

- Androids now have a charge, similar to Ethereals. 
   - Like Ethereals...
      - ...They can drain charge from APCs and such like Ethereals
      - ...They can overcharge
   - But unlike Ethereals
      - ...They cannot consume Liquid Electricity to recharge
      - ...Food and Drink(alcohol) they consume is converted to power
      - ...Running out of power is lethal
      - ...Power drains more actively than Ethereals
      - ...Low power reduces consciousness directly rather than cause toxin damage (subject to change)

- Androids can taste, but don't like or dislike any food or drink type. They also can't be disgusted.

- Androids have unique mechanics around poisons,
   - Androids retain their immunity to a majority of chemicals, but poisons will build up a "toxicity" stat representing clogged filters, which causes penalties to consciousness and movespeed. (subject to change) 

- Androids passively produce heat. This is increased with high toxicity or decreased with low power.

- Temperature affects Androids, but slightly differently. 
    - Heat and Cold do direct damage, rather than building Hypo or Hyperthermia. They will also cause action and movespeed slowdown.
    - They are affected by cold significantly less than by heat.
    - Cold makes them more brittle, taking more damage from blunt attacks.

- Androids breathe, but to maintain their body temperature rather than to use gas like Oxygen. 
   - Any negative effects of gases don't affect them. What they breathe doesn't matter, it's how hot it is that matters. This means "breathing in" a plasmafire will lead you to overheat. 

- Android brains are immune to minor level traumas. 

- Androids are less affected from lacking blood (oil) but do not passively regenerate blood. 

- Androids can select a species to base their model off of in character customization. Some species have unique customization in the body menu. 

- Androids can pick a lawset from a small selection. Unlike silicons, these laws are effectively immutable outside of ion storms, and mostly serve as voluntary roleplay guidelines. They even get a law 0 if they're an antag!

- Androids can pick to be emotionless, disabling all positive and negative moodlets. 

#### Future changes

- Removed radiation immunity, replacing it with warping of wires and such.
- Inability to defibrillate, replace with a unique revival method.
- More variety in EMP effects. 
- Androids can bleed (oil)

### Synthetics

- Synthetics remain distinct from Androids, but now descend from them, meaning they share the new mechanics of Androids while also being disguised. 

### Other

- Some misc changes to rendering
- Organ repair surgeries can be performed infinitely on unremovable organs


CRITAWAKETS:
this seems really fun but i have two things im not really all for
- lack of food causing death is just not fun. in both gameplay and roleplay it just sucks. might be worth looking into a way less punishing penalty for ignoring electricity intake like massively lowering movespeed, depleting all run energy and overall just entering a "low power" mode.

- Organs being irremovable on androids seems like a waste considering robotics like, exists. At the very least, they should be removable and have new organs printable from robotics even without upgrades being a thing. (in the future i would like any new organ replacements for androids to be more like very impactful sidegrades than any sort of upgrade)

Constellado:
I will be the main test subject of this PR, for sure! I will put more comments after the next round this is in. I will be playing as a synth for this. I do like the power mechanic, it did feel very strange just being able to do so much without having to stop to eat or do something.

I do think there should be a part whitelist to replace the robot parts with, like you can only put in robot body parts of any kind but not organic ones, maybe?.

I do want synths to have normal humanoid voices, too! I don't know if this is getting fixed here. If it is, great!

The ability for roboticists or people in med to replace the battery if you "die" as a synth I think is important! There should be something more than putting a synth in a suit charger. 

I have a feeling I am going to forget to charge. Can there be red messages in chat to remind me to charge?

Think you can add the ability to use an inducer on yourself and other synths?

Constellado:
I have liked the changes! Being low on power is definitely noticeable and is not hard to manage. I personally would like the ability to write my own laws but putting no laws is fine enough for me :D

PR #895, opened 2026-03-11
TGUI Spawn Panel, TGUI Spawn, fixed TGUI Nuke Pane, abstract_type moved to /datum

CRITAWAKETS:
image

And much more.

Good luck test-merging this as the abstract types PR will conflict with basically everything that isn't completely modular.


MrMelbert:
Fails all tests because of map icons, need to recompile with story content enabled probably

PR #934, opened 2026-08-07
Adds a new UI option

Constellado:
# What is this?
I wanted a super dark mode UI, named Carbon.
It is also canon to what my character's prefer; **Dark** mode. The pitch black background theoretically uses less electricity with OLED screens!
Its sorta ugly, in a way. I may remove the black borders around the UI? unsure.
Also, there is engi themed inventory icons instead of sec themed!

I also noticed there is a wider gap between the ID and the belt icons, so I may fix that.

# Screenshot!
image
image


Constellado:
I may make the icons a tad darker compared to the borders if it ends up burning my eyes too much in a round.

Constellado:
Updated it so it is less eye burning (and easier to tell if something is active!) (ignore the changeling clothes I spawned with that somehow)
image
Act sleep didn't work even though there is an icon for it, not too sure what's going on there.
image

